PR #3, opened 2024-02-01
Ct/typescript linting

Craiting:
Adding a linter and starting to fix types. Fixing all types will be time consuming. I was able to get the process started but plan to come back and fix them all later. The server runs as long as "strictNullChecks": false in the tsconfig.json. In the future we should set that field to true and fix the type errors.

The server runs for now.


emilysallstrom (approved):

tphalp (approved):
:shipit:

PR #4, opened 2024-02-02
Update readme

Craiting:
Adding some readme instructions for what I think the ucw-app could become. A lot of this could change, the current ucw-app doesn't actually work how the readme says it should.


README.md
```diff
@@ -1 +1,35 @@
-# ucw-app
\ No newline at end of file
+# Universal Connect Widget (Application)
+
+This repo is a full stack application which anyone can clone and self-host as a way to serve the connect widget via a url which can then be loaded into an iframe.
```
tphalp, line 3: My only question here is whether or not the final widget package will serve a url. If we go the route of exposing a React Component, then it won't work like this. 

If we go the route of using docker to serve the widget url, or use `single-spa.js` to serve the frontend, then it could continue to work as it originally did.

README.md
```diff
@@ -1 +1,35 @@
...
+
+#### Getting Started (in development)
+1. clone the repo
+2. copy .env.example
+3. Follow Initial setup for getting environment variables
+4. run `docker-compose up`
+5. load a widget url `http://localhost:8080/?job_type=agg&auth=${authCode}`
```
emilysallstrom, line 10: I wonder if # 5-7 would make more sense in documentation about the server - something like APIs available through the server. Or maybe point to the example app for how to use `ucw-app`. I don't think they're technically needed to "Get started", but more of a how-to.

README.md
```diff
@@ -1 +1,35 @@
...
+DO NOT put any credentials in any of the js files. If you do so, it could accidentally get committed and leaked to public.
+USE `.env` FILE
+
+*UCP credentials are required for authentication and secret exchange, storage (redis-like session cache) and analytics services.*
+
+*The `CryptoKey` and `CryptoIv` values are for encrypting the session token in order to not rely on cookies. They must be shared across server instances if there are multiple instances.*
+
+* You might see an error about failure to connect redis, the widget doesn't rely on redis to start, but some providers logic require an redis intance, to fix this error you can either: 
+- start a local redis instance, this way it will be avaliable at localhost:6379 and the widget will use it
+- Or set in `.env` Env=dev, this way the redis client will use local in-mem object to handle the cache and remove the error, however, this is just for some testing, the cached values won't expire and also will be cleared on server restart.
```
emilysallstrom, line 35: This might be a good thing to add to `.env.example`

.env.example
```diff
@@ -0,0 +1,6 @@
+CryptoKey=""
+CryptoIv=""
+
+UcpAuthClientId=""
+UcpAuthClientSecret=""
+UcpAuthEncryptionKey=""
```
emilysallstrom, line 6: Thank you for adding this!!

README.md
Craiting: I think either way it'll load from a url because the way it's implemented currently is to hit the server url with an authCode and the server loads the front-end widget. So whether it's loaded via a microfrontend or react component it will still have to be imported into the front-end part of the full stack ucw-app. I think the decision to use react or a microfrontend only really impacts users who want to use the widget alone in their app and build their own backend for it.

README.md
Craiting: Yeah good point. I'll update this.

emilysallstrom (approved):

PR #6, opened 2024-02-21
Fix existing lint errors

tphalp:
This PR adds additional eslint configs, and fixes all linting errors.


.eslintrc.js
```diff
@@ -0,0 +1,26 @@
+module.exports = {
+  root: true,
+  env: {
+    'browser': true,
+    'node': true,
+    'es6': true,
+    'jest': true
+  },
+  extends: [
+    'eslint:recommended'
```
Craiting, line 10: I wonder if it'd be better in the long run to use `standard-with-typescript`? It seems to check the typescript files a bit better. I just asked chatgpt what's the difference and it seems like if your using typescript eslint:recommend doesn't really do typescript linting.

> `eslint:recommended` and `standard-with-typescript` are both popular ESLint configurations, but they serve slightly different purposes:
> 
> 1. **eslint:recommended**:
>    - `eslint:recommended` is a built-in ESLint configuration that includes a set of rules recommended by the ESLint team.
>    - These rules are considered as a good starting point for many projects and aim to catch common errors and stylistic issues.
>    - `eslint:recommended` focuses on general JavaScript best practices and does not include rules specific to TypeScript.
> 
> 2. **standard-with-typescript**:
>    - `standard-with-typescript` is a configuration based on the StandardJS style guide but extended to support TypeScript.
>    - StandardJS is a popular style guide and ESLint configuration that promotes a consistent coding style and emphasizes simplicity and readability.
>    - The `standard-with-typescript` configuration includes rules for both JavaScript and TypeScript, making it suitable for projects using TypeScript.
>    - It enforces rules for TypeScript-specific syntax and conventions in addition to the standard JavaScript rules.
> 
> In summary, while `eslint:recommended` provides a basic set of rules for JavaScript, `standard-with-typescript` extends those rules to support TypeScript, adhering to the StandardJS style guide along with TypeScript-specific conventions. Your choice between the two would depend on whether you're working with TypeScript and if you prefer the StandardJS style guide.

tphalp: Okay, thanks for this Craig. I'll do some testing with it.

tphalp: And, nice usage of ChatGPT 😁

tphalp: Ha, when I switch from `eslint:recommended` to `standard-with-typescript`, I go from 0 errors, to 2250 errors. 😮‍💨

tphalp:
@Craiting When I add TS lint rules, there are a TON more errors....I think I'll do the TS stuff in a different branch.

emilysallstrom:
Agreed - that does look like the bettor solution, but let's address those errors in a future PR.

emilysallstrom (approved):

PR #9, opened 2024-02-23
Add gh actions file for test

tphalp:


package.json
```diff
@@ -70,6 +70,7 @@
     "eslint-plugin-promise": "^6.1.1",
     "husky": "^9.0.11",
     "lint-staged": "^15.2.2",
+    "msw": "^2.2.1",
```
Craiting: I see you added msw but I don't see it being imported anywhere, was this accidental?

tphalp: I removed it. That was an accident.

emilysallstrom (approved):

PR #12, opened 2024-02-28
ts fixes and change requires to import

Craiting:
Updating the code to more modern Javascript. I tested mx bank and sophtron bank and things are still working. There are still a ton of typescript errors and we need to turn on `strictNullChecks` but once that's turned on we will need to fix more errors.


tphalp (approved):
:shipit: Thanks for doing this, Craig.

PR #19, opened 2024-03-14
Add Widget NPM component AND Dockerfile updates

tphalp:
This adds the UI that contains the UCW component from our npm module. It is served by the `server` service, just like before, but instead of the UCW code living in this repo, it is being consumed from the NPM package.

A new folder, `ui`, has been added, which contains a small React/Vite app that serves the widget component. Look at the `ui/README.md` file for more info.

New commit builds off of the NPM work, and also adds an updated Dockerfile, so that build/run works as expected, as well as the ground-work for docker compose. Right now there is a bug in docker compose when reading the env file. I will address that in another PR


Dockerfile
```diff
@@ -1,21 +1,41 @@
...
-RUN sed -i '35s/.*/return try JSONSerialization.jsonObject(with: data, options: [.mutableContainers, .fragmentsAllowed])/' /app/node_modules/@capacitor-community/http/ios/Plugin/HttpRequestHandler.swift
-    
-RUN npm install -g nodemon ts-node
+# Test 2
+RUN cd /app/ui &&  \
+    npm run build
+# End test 2
+
+## UI
+#RUN cd ./ui &&  \
```
WesRisenmay, line 31: Can we get rid of this?

package.json
```diff
@@ -1,9 +1,10 @@
 {
   "name": "ucw-app",
   "version": "0.1.0",
-  "private": true,
   "scripts": {
-    "server": "npx ts-node server/server.js",
+    "server": "ts-node server/server.js",
+    "ui": "cd ./ui && npm run dev",
```
WesRisenmay, line 6: This feels brittle. Wouldn't this fail if you ran it twice, because you would have already changed folders?

server/server.js
```diff
@@ -66,35 +66,36 @@ function renderDefaultPage (req, res, html) {
...
       res.sendStatus(404)
     }
   })
 } else {
-  info('using local resources from "../build"')
+  info('using local resources from "../ui/dist"')
   app.get('/', async (req, res) => {
-    const filePath = join(__dirname, '../', 'build', 'index.html')
+    const filePath = join(__dirname, '../ui', 'dist', 'index.html')
+    console.log('filePath', filePath)
```
WesRisenmay, line 89: Do we still need this log?

server/server.js
```diff
@@ -66,35 +66,36 @@ function renderDefaultPage (req, res, html) {
...
     const html = await readFile(filePath)
     renderDefaultPage(req, res, html)
   })
-  app.get('*', _static(join(__dirname, '../build')))
+  app.get('*', _static(join(__dirname, '../ui/dist')))
 }
 
+// app.listen(config.Port, config.Env === 'prod' ? '0.0.0.0' : 'localhost', () => {
```
WesRisenmay, line 96: Can we get rid of this?

server/server.js
```diff
@@ -66,35 +66,36 @@ function renderDefaultPage (req, res, html) {
...
 app.listen(config.Port, () => {
-  const message = `Server is running on port ${config.Port}, env: ${config.Env}`
+  const message = `Server is running on port ${config.Port}, env: ${config.Env}\n`
+  console.log(message)
```
WesRisenmay, line 99: Do we need both an info and a console.log?

ui/src/assets/react.svg
```diff
@@ -0,0 +1 @@
...2.6a321.187 321.187 0 0 0 7.257 13.246c2.594 4.48 5.28 8.868 8.038 13.147Zm37.542 31.03c-5.184-5.592-10.354-11.779-15.403-18.433c4.902.192 9.899.29 14.978.29c5.218 0 10.376-.117 15.453-.343c-4.985 6.774-10.018 12.97-15.028 18.486Zm52.198-57.817c3.422 7.8 6.306 15.345 8.596 22.52c-7.422 1.694-15.436 3.058-23.88 4.071a382.417 382.417 0 0 0 7.859-13.026a347.403 347.403 0 0 0 7.425-13.565Zm-16.898 8.101a358.557 358.557 0 0 1-12.281 19.815a329.4 329.4 0 0 1-23.444.823c-7.967 0-15.716-.248-23.178-.732a310.202 310.202 0 0 1-12.513-19.846h.001a307.41 307.41 0 0 1-10.923-20.627a310.278 310.278 0 0 1 10.89-20.637l-.001.001a307.318 307.318 0 0 1 12.413-19.761c7.613-.576 15.42-.876 23.31-.876H128c7.926 0 15.743.303 23.354.883a329.357 329.357 0 0 1 12.335 19.695a358.489 358.489 0 0 1 11.036 20.54a329.472 329.472 0 0 1-11 20.722Zm22.56-122.124c8.572 4.944 11.906 24.881 6.52 51.026c-.344 1.668-.73 3.367-1.15 5.09c-10.622-2.452-22.155-4.275-34.23-5.408c-7.034-10.017-14.323-19.124-21.64-27.008a160.789 160.789 0 0 1 5.888-5.4c18.9-16.447 36.564-22.941 44.612-18.3ZM128 90.808c12.625 0 22.86 10.235 22.86 22.86s-10.235 22.86-22.86 22.86s-22.86-10.235-22.86-22.86s10.235-22.86 22.86-22.86Z"></path></svg>
```
WesRisenmay, line 1: I think you can get rid of this, and it's not being used. Am I crazy?

ui/src/vite-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="vite/client" />
```
WesRisenmay, line 1: Do we need this comment?

package.json
tphalp: Yeah, good point. I added this for use in the Dockerfile, but I'll figure out a different way to do this.

server/server.js
tphalp: I added this so it is clearer that the server is running, and on which port, when you run this from the terminal. The original output shows this info, but it is buried in an object, and is easy to miss, especially for someone new to the project. However, if this is bad form, I'll remove it.

ui/src/vite-env.d.ts
tphalp: This was created by Vite automatically, so I'm not entirely sure. I'll look into it.

server/server.js
WesRisenmay: Sounds like it's worth keeping it

package.json
tphalp: I ended up splitting the UI app into its own docker container, and was able to get rid of this.

ui/src/vite-env.d.ts
tphalp: I haven't looked into this, but I will in another PR.

PR #21, opened 2024-03-25
refactored config to use mostly env variables. documented where varia…

WesRisenmay:
…bles come from and what they are


.env.example
```diff
@@ -1,7 +1,27 @@
...
+HostUrl="https://test.universalconnectproject.org"
+WebhookHostUrl="https://webhook.universalconnectproject.org"
+
+LogLevel="debug"
+
+Env="pre" #the mocked environment will mock out your https calls as defined in mock.js
+
+RedisServer="redis://localhost:6379"
+RedisCacheTimeSeconds="600"
+ResourcePrefix="http://127.0.0.1:3000" #http://localhost:5173 for docker, or http://127.0.0.1:3000 if running locally
```
emilysallstrom, line 26: Do you know what "resource" is referring to here? I'm wondering if this will change based on @tphalp's recent changes. If so, we can probably leave it as-is. If not, we may want to add a description for resource. (I have no idea what it's referring to 😆 )

WesRisenmay: I think it's where the server is listening for requests. Is that right @tphalp ?

tphalp: Yes, `ResourcePrefix` is where the UI lives (connect widget). This really shouldn't be changed, if they are using the docker images. If they end up putting their own infrastructure up, then changing it would make sense.

tphalp: That really not a great name. I'm open to renaming it, so it makes more sense.

tphalp: Oh, I just re-read your question, @WesRisenmay  

It's not where the server itself listens for requests. That is just `localhost`, and the port listed in config, `Port: '8080'`.

`ResourcePrefix` is the url that the server uses to serve the Widget UI.

tphalp: In my new docker changes, this actually gets changed in the docker-compose file:

https://github.com/Universal-Connect-Project/ucw-app/blob/main/docker-compose.yaml#L16-L18

We'll probably want to call that out in the readme. I hadn't thought of the impact of that. 

Honestly, if they are using the all-in-one docker stuff, and not piece-mealing their own solution (taking each piece, and spinning up their own instance), then it really shouldn't affect them.

tphalp: Go ahead and merge this. I'll mess around with making the docker compose file use the env file

tphalp (approved):
:shipit:

PR #23, opened 2024-03-26
Better Docker Patterns

tphalp:
This PR cleans up the docker files, docker compose file, deletes the local docker scripts, and adds the new docker info to the READMEs. 

The original Docker scripts will not work any more as stand-alone containers, due to the changes in how the two systems (server, and widget) now work together. The docker-compose config makes it much easier, and gives the same result.

With the changes to the Dockerfiles, I was also able to get the docker images down from a combined 2.1GB, to under 800MB.


Craiting (approved):
Nice!

PR #24, opened 2024-03-29
Updates to allow the .env to work with docker compose; other clean-up

tphalp:
- Moves `ts-node` from deps to devdeps
- Uses `serve` instead of `vite preview` to serve the ui. 
- Uses the `.env` for docker compose, so that the correct ports are used across the docker ecosystem. 
- Fixes an issue where the health checks were leaving files on the docker filesystem.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,5 +23,8 @@ Env="pre" #the mocked environment will mock out your https calls as defined in m
 
 RedisServer="redis://localhost:6379"
 RedisCacheTimeSeconds="600"
+RedisPort="6379"
+
 ResourcePrefix="http://localhost:5173" #the url that serves the widget ui
 ResourceVersion=""
+ResourcePort="5173"
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,11 +4,10 @@
 FROM alpine:3.19.1
 ENV NODE_VERSION 20.11.1
 
-RUN apk --update --no-cache add nodejs npm wget
+RUN apk --update --no-cache add nodejs npm
 
 WORKDIR /app
 
-# Server
 COPY package.json package-lock.json tsconfig.json tsconfig.paths.json /app/
 RUN npm ci --only=production &&  \
     npm i -g ts-node
@@ -17,9 +16,7 @@ COPY ./server ./server
 COPY ./shared ./shared
 
 ENV Env prod
-ENV Port 8080
-ENV ResourcePrefix 'http://localhost:5173'
 
-EXPOSE 8080 6379
+EXPOSE ${Port}
 
 CMD ["ts-node", "server/server.js"]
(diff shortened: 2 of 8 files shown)


.env.example
```diff
@@ -23,5 +23,8 @@ Env="pre" #the mocked environment will mock out your https calls as defined in m
 
 RedisServer="redis://localhost:6379"
 RedisCacheTimeSeconds="600"
+RedisPort="6379"
+
 ResourcePrefix="http://localhost:5173" #the url that serves the widget ui
 ResourceVersion=""
+ResourcePort="5173"
```
tphalp, line 30: I added an env specifically for the port, because the docker compose file uses different conventions for specifying the host (not localhost).

Dockerfile
```diff
@@ -4,11 +4,10 @@
 FROM alpine:3.19.1
 ENV NODE_VERSION 20.11.1
 
-RUN apk --update --no-cache add nodejs npm wget
+RUN apk --update --no-cache add nodejs npm
```
tphalp, line 7: `wget` is already on the image

docker-compose.yaml
```diff
@@ -7,7 +7,7 @@ services:
     container_name: ucw-app-server
     restart: on-failure
     healthcheck:
-      test: ["CMD", "wget", "http://server:8080/ping"]
+      test: ["CMD", "wget", "--no-verbose", "--tries=1", "--spider", "http://server:${Port}/ping"]
```
tphalp, line 10: There was an issue with wget leaving temp files in the container file system

package.json
```diff
@@ -71,6 +70,7 @@
     "lint-staged": "^15.2.2",
     "prettier": "^3.2.5",
     "ts-jest": "^29.1.2",
+    "ts-node": "^10.9.2",
```
tphalp, line 73: Move this to devdeps, instead of deps.

ui/Dockerfile
```diff
@@ -16,16 +16,16 @@ RUN npm run build
 FROM alpine:3.19.1 as web
 ENV NODE_VERSION 20.11.1
 
-RUN apk --update --no-cache add nodejs npm wget
+RUN apk --update --no-cache add nodejs npm
```
tphalp, line 19: `wget` already available

ui/Dockerfile
```diff
@@ -16,16 +16,16 @@ RUN npm run build
...
 
 WORKDIR /app
 
 COPY package.json package-lock.json vite.config.ts /app/
 RUN npm ci --only=production &&  \
-    npm i -g vite@5.2.6
+    npm i -g serve@14.2.1
```
tphalp, line 25: Use `serve` instead of `vite`

emilysallstrom (approved):

PR #26, opened 2024-04-01
Fix var name

tphalp:
I pulled and ran the code for the server, and it was throwing an error because of this variable name.


Craiting (approved):
Oh thanks for fixing that

PR #27, opened 2024-04-02
Update readme for better clarity, and add combined NPM script to run …

tphalp:
…both together

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ To get started, clone the repo, follow the steps in `Initial Setup`, below to se
 
 *This assumes you have [docker](https://docs.docker.com/get-docker/) and [docker-compose](https://docs.docker.com/compose/install/) already installed.*
 ```
-docker compose --env-file .env up
+docker compose up
 ```
 
 That's it! If you have questions, please reach out to us.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -10,7 +10,7 @@ To get started, clone the repo, follow the steps in `Initial Setup`, below to se
 
 *This assumes you have [docker](https://docs.docker.com/get-docker/) and [docker-compose](https://docs.docker.com/compose/install/) already installed.*
 ```
-docker compose --env-file .env up
+docker compose up
```
tphalp, line 13: The explicit env-file param isn't necessary, since we're referencing the env file in docker-compose.yaml

package.json
```diff
@@ -4,6 +4,8 @@
   "scripts": {
     "server": "ts-node server/server.js",
     "server:debug": "node --inspect -r ts-node/register ./server/server.js",
+    "widget": "cd ./ui && npm run dev",
```
tphalp, line 7: I tested this, and even if you call `npm run widget` multiple times in a row, calling `cd ./ui` doesn't cause issues. It is idempotent, and thus doesn't seem to be fragile.

package.json
```diff
@@ -4,6 +4,8 @@
...
+    "combined": "npx --yes npm-run-all --parallel server widget",
```
emilysallstrom, line 8: My only thought is that using "combined" as the command means you need to really understand what's happening behind the scenes. What about `ucw` or `ucw-app`? If I were a developer who didn't read the docs, and I saw these options, it may be more intuitive than "combined". Just a thought, I could be wrong 🤷

emilysallstrom (approved):

package.json
tphalp: Good call-out. I'll change it.

emilysallstrom (approved):

emilysallstrom:
Thanks Tyson!

PR #30, opened 2024-04-08
Wesr/cypress

wesrisenmay-mx:


cypress/support/providerCredentials.ts
```diff
@@ -0,0 +1,60 @@
+const providerCredentials = {
+  mxInt: {
+    username: Cypress.env('mx_client_id'),
+    password: Cypress.env('mx_api_secret'),
+    basePath: 'https://int-api.mx.com',
+    vcEndpoint: 'https://int-api.mx.com/',
```
Craiting, line 6: I know the code has a few places already where url variables are inconsistent whether they end with a "/" or not. In my opinion none should end in a slash. But it would be nice to establish a pattern and stick to it. What do you think?

cypress/support/providerCredentials.ts
```diff
@@ -0,0 +1,60 @@
...
+    provider: 'mx_int',
+    available: true
+  },
+  mxProd: {
+    username: Cypress.env('mx_client_id_prod'),
+    password: Cypress.env('mx_api_secret_prod'),
+    basePath: 'https://api.mx.com',
+    vcEndpoint: 'https://api.mx.com/',
```
Craiting, line 14: Another one.

cypress/support/commands.ts
```diff
@@ -0,0 +1,74 @@
+import '@testing-library/cypress/add-commands'
```
tphalp, line 1: This is super cool. I didn't know you could do this in Cypress.

tphalp (approved):
:shipit:

Craiting (approved):

PR #35, opened 2024-04-12
Jishun/localauth

sophtron-jishun:
carried over from https://github.com/Universal-Connect-Project/universal-connect-widget/pull/57


server/serviceClients/authClient/index.js
```diff
@@ -1 +1,17 @@
-export * from './ucp'
+import {AuthClient as UcpAuth} from './ucp'
+import {AuthClient as SophtronAuth} from './ucp'
+import {AuthClient as LocalAuth} from './local'
+const config = require('../../config');
```
wesrisenmay-mx, line 4: Why not import instead of require?

server/serviceClients/authClient/local.js
```diff
@@ -0,0 +1,83 @@
+// This is an example of the simplest scenario when the widget is hosted and used in the same secure domain
+// where as the provider credentials are safe to be stored in the widget server, hence as soon as the auth token is valid
+// Or no auth token needed at all, 
+// simply return the provider credentials, coppied from : 
+// const providerConfig = require('../../../../example/application/configuration')
+
+const {decrypt} = require('../../utils');
```
wesrisenmay-mx, line 7: Why not import here as well?

server/serviceClients/authClient/local.js
```diff
@@ -0,0 +1,83 @@
...
+
+  async getSecretExchange(iv){
+
+    const key = Buffer.from(config.LocalAuthEncryptionKey, 'base64').toString('hex')
+    const token = decrypt(this.token, key, iv);
+    // if token is valid, or use your own way to validate the token.
+    if(token === config.LocalAuthPhrase){ 
+      return {
+        mxInt: {
+          username: process.env.MxClientId,
```
wesrisenmay-mx, line 24: We should probably import these from config.js, so we have a clear place that tells us what we're pulling from .env

server/serviceClients/authClient/local.js
```diff
@@ -0,0 +1,83 @@
...
+          password: process.env.MxApiSecret,
+          basePath: 'https://int-api.mx.com',
+          vcEndpoint: 'https://int-api.mx.com/',
+          provider: 'mx_int',
+          available: true
+        },
+        mxProd: {
+          username: process.env.MxClientIdProd,
```
wesrisenmay-mx, line 32: It looks like you added several variables to the .env file. These should be added to .env.example. It may make sense to group the things that are only needed with local auth together under a comment describing when you would actually need them

server/config.ts
```diff
@@ -15,30 +15,36 @@ const nonSensitiveSharedConfig = {
...
-  'Env',
-
-  'RedisServer',
-  'RedisCacheTimeSeconds',
-  'ResourcePrefix',
-  'ResourceVersion'
+  "CryptoKey",
+  "CryptoIv",
+
+  "AuthProvider", //supported: ucp, local, sophtron
```
wesrisenmay-mx, line 21: We've been putting comments describing these variables in the .env.example file instead of here

server/config.ts
```diff
@@ -15,30 +15,36 @@ const nonSensitiveSharedConfig = {
...
+  // fill the corresponding credentials below if the auth provider is to be used
+
+  "LocalAuthEncryptionKey", // shared encryption key with caller to encrypt the auth phraze
```
wesrisenmay-mx, line 24: It seems like we need something added to the readme describing when you would want to use these and how to set them up

server/serviceClients/authClient/index.js
```diff
@@ -1 +1,17 @@
...
+
+export function CreateAuthClient(token, authProvider){
```
wesrisenmay-mx, line 6: This function needs a unit test

server/serviceClients/authClient/index.js
```diff
@@ -1 +1,17 @@
-export * from './ucp'
+import {AuthClient as UcpAuth} from './ucp'
```
wesrisenmay-mx, line 1: It's confusing to import the same thing and name it two different things. This had me looking in the UCP file wondering what was different about UcpAuth vs SophtronAuth

server/serviceClients/authClient/index.js
```diff
@@ -1 +1,17 @@
...
+import {AuthClient as SophtronAuth} from './ucp'
+import {AuthClient as LocalAuth} from './local'
+const config = require('../../config');
+
+export function CreateAuthClient(token, authProvider){
+  switch((config.AuthProvider || authProvider).toLowerCase() ){
+    case 'sophtron':
```
wesrisenmay-mx, line 8: Why does this sophtron case need to exist if it's functionality doesn't differ from ucp?

server/serviceClients/authClient/local.js
```diff
@@ -0,0 +1,83 @@
...
+const {decrypt} = require('../../utils');
+const config = require('../../config');
+
+export class AuthClient {
+  token;
+  constructor(token){
+    this.token = token;
+  }
+
+  async getSecretExchange(iv){
```
wesrisenmay-mx, line 16: This function needs a unit test

server/config.ts
wesrisenmay-mx, line 21: We're going to need an end to end test that makes sure our auth is working when using the local AuthProvider.

You'll need to add some secrets and environment variables to this repo. If you don't have access, then ping @emilysallstrom.

You'll also need to add an extra github action that runs the e2e tests with the local authprovider. For right now it can just rerun all the tests with local auth. In the future we might want to only run a subset of the e2e tests with the local provider.

server/serviceClients/authClient/index.js
sophtron-jishun: they are not the same thing, they are 2 non-related providers that happens to look similar

server/serviceClients/authClient/local.js
sophtron-jishun: please read the comments at the top, these are not going into the config because these are not the standard config.  it's not intended to let clients use this provider as is. but we can't provide something that's good enough for anyone, because each clients will have their own standard auth process, whoever whats to use this, have to come to this file, fullly understand it and make their customisations

server/serviceClients/authClient/index.js
sophtron-jishun: they are not the same thing, they are 2 non-related providers that happens to look similar

server/serviceClients/authClient/index.js
Craiting: I think you meant to import from sophtron
```javascript
import { AuthClient as SophtronAuth } from './sophtron'
```

server/serviceClients/authClient/index.js
wesrisenmay-mx: As Craig mentioned above it looks like the import is wrong. Right now they are the same thing, but after changing the import they won't be the same thing

server/serviceClients/authClient/local.js
wesrisenmay-mx: I guess I'm confused about why we are developing this feature if we don't expect anyone to use it.

If we're going to support the use case of not using an auth server and storing credentials within the app server, then it seems like we should make it obvious of how to do that by making example env files and readme descriptions of how to do this. We'll also need end to end tests to make sure it works.

If we're not planning to support this use case, then I don't understand why we're adding it into the code.

server/serviceClients/authClient/index.js
sophtron-jishun: this is fixed

server/serviceClients/authClient/local.js
sophtron-jishun: you need to understand the whole design of the widget , part of the background of this is here https://sophtron.slack.com/archives/C06FK8V1BFW/p1711558034510199
it's said at the top, an example for customers to customize

sophtron-jishun: please see readme changes in the original PR, 
https://github.com/Universal-Connect-Project/universal-connect-widget/pull/57/files#diff-679a7f928b73ed9c06f8374dbe256645697332a761d6f6ea7f1b87cfc3a6271d
there is no suitable place for this in this repo, @tphalp mentioned it's being worked on

sophtron-jishun: tagged you in the thread,

wesrisenmay-mx: I believe I understand how the 2 different forms of authentication work. I don't understand why we wouldn't want to put some documentation into our readme and example env files to support both forms of authentication

sophtron-jishun: I explained that, the proper position in the readme about this is not available and @tphalp is working on it

sophtron-jishun: the comments in config were moved to `.env.example`

tphalp: @sophtron-jishun Can you remind me what is still being worked on? I can't remember what we talked about.

Craiting (commented):
I think this local auth option could become the default recommended implementation option for clients going the self-hosted route. It seems like the majority of use cases will be self-hosted. It makes sense to me to add the new config variables in the example.env just in it's own section, then a client trying to set it up for self-hosted can easily fill in those variables without needing to copy and paste variables from the readme. If the client is not doing self hosted they can ignore the variables or delete them. 

In general I think we want to build an easy "turn-key" solution which will work for the majority of clients and then allow clients that the turn-key solution doesn't work for to customize it to suit their needs. We can provide documentation and examples for different implementation options and it's expected that those clients who require a lot of customization will need to really understand what's happening in the code and between all the different services. But for the basic client who just wants to get something up and running as quickly as possible, I don't think they should have to understand everything in the code, they should just be able to follow a quick step by step guide to configure the variables to get UCW up and running as quickly as possible.

server/config.ts
```diff
@@ -15,30 +15,35 @@ const nonSensitiveSharedConfig = {
...
+  "HostUrl",
+  "WebhookHostUrl",
+  "LogLevel",
+  
+  "Env",
+
+  "RedisServer",
+  "RedisCacheTimeSeconds",
+  "ResourcePrefix",
+  "ResourceVersion",
```
Craiting, line 46: We're using `standard-with-typescript` styles which prefers single quotes over double quotes. Please install a linter in your IDE to make these style requirements more obvious as you're coding. Eventually, when we get all the typescript errors resolved there will be a pipeline to check for linting errors and block pull requests. But we're trying to follow common patterns to make this repo more uniform and easy to read for future devs who will be reading or contributing to the code.

server/serviceClients/authClient/local.js
Craiting: The goal for this repo is to have complete test coverage to ensure confidence whenever we make a change. We're a long way from this goal but it would be really helpful to start by making sure at least the new code that's added is covered by tests. Wes has been working on adding some tests to provide a good example which we can follow when writing new tests. Have those been merged yet @wesrisenmay-mx? Can you share a link to one?

wesrisenmay-mx: Unit/Integration test example: https://github.com/Universal-Connect-Project/ucw-app/blob/main/server/providers/mx.test.ts

End to end test examples:
https://github.com/Universal-Connect-Project/ucw-app/blob/main/cypress/e2e/mx.cy.ts

https://github.com/Universal-Connect-Project/ucw-app/blob/main/cypress/e2e/sophtron.cy.ts

sophtron-jishun:
@Craiting the `LocalAuth` provider is an example and so far as I can think of, this will stay as an example for longer unless we find a "generic" way to handle "any" clients' own auth requirement,  in which case , I don't think it's yet possible, 

imagine if you are handling user auth for your org, if it's a big company, it should have its own identity/auth/security service with their own requirements, we can't anticipate how it will look like, most likely each one who comes up will be a different one. 

a small client would more likely want to use "the simple way", however, we can't let them to use this either, without introducing another persistency layer,  this will be only a static configured phrase, which is not going to be secure enough, so that I don't think I will let my clients to use such a "thing" as is. they'll need to choose either to do the homework and integrate with their auth provider, as the big companies do. or they should choose to use the "UCP" auth we provide, which is kinda the minimal secure implementation. 

therefore, I agree that we should make this the recommended way for our clients. but as above, we can't make it default, a default way that's minimal and secure enough in our UCP context would be the UCP auth. 

furthermore, if we do want to provide clients something out of box with a "simple" flow, we may choose to provide more implementations of the local auth provider, such as,  integrating with frameworks like `identityServer`, or popular providers like `okta` `auth0` etc , in either case, I think adding those lot of credential configurations into the config is a lot of noises,

server/serviceClients/authClient/local.js
tphalp: Yes, what I was referring to is the docs. Our docs team is still working on them. I imagine we will eventually host them on the UCP website.

Craiting:
@sophtron-jishun that makes sense to me. We discussed this a bit and I think the general agreement from the MX team is that maybe we should postpone this feature until we have more time to make it into something that we feel is secure and we would feel confident recommending clients to use. Considering the long term it seems like it might be a good idea to include documentation and examples of different use cases, where something like this could potentially live in the future. We're hoping with ucw-app to build something that will work out of the box for most clients, so we'd like to limit complexity and reduce the learning curve for new devs coming into the app by offering a recommended solution for the most common use case. It sounds like local auth currently would require some special configuration and it would not be considered secure or recommended so it might be best to just leave it out to prevent new devs coming in from getting confused. Because they might not initially know which parts of the code are meant to be used and which are not.

Anyway, what do you think about closing this and putting it off for a later date? I think the idea of local auth is good but I wonder if there is more we could do to make it secure and recommended. We would like to revisit this but our current priorities are getting comprehensive tests and making sure the code is working for each provider.

tphalp:
Direction has changed. Closing.

PR #41, opened 2024-04-22
added a monorepo ADR

wesrisenmay-mx:


architectureDecisionRecords/0003-monorepo.md
```diff
@@ -0,0 +1,31 @@
...
+
+Each service will have its own .env file.
+
+## Consequences
+
+Contributors will need to have a basic working knowledge of turborepo to contribute.
+
+Making changes will be easier, because you won't need to put PRs up for multiple repositories when they need changes to be able to work together.
+
+Shared packages will be easy to create, and they won't require publishing to npm.
```
tphalp, line 29: One point that Emily pointed out, is that the widget needs to be retained as a separate package so that companies that want to customize the widget won't need to do so from the monorepo. MX's use case is an example of this. 

Because of this, I'm not sure of the best way to incorporate the NPM widget repo into this scenario.

wesrisenmay-mx: We discussed a bit in standup, but not all shared packages have to live here. I don't think the widget should live here, but there will be other small packages of shared code that should.

tphalp: One thing that Jishun has brought up, is that the way we use the NPM package currently is pretty cumbersome for customization, and even for updates. Maybe we can brainstorm how that can be improved.

Craiting:
I like the idea of a monorepo because it seems like the easiest way to allow developers to self host their own UCW service.

tphalp (approved):
I like this idea. Having used a monorepo before, I'm hopeful that this will give us a cleaner way to develop, and an easier way for our end-users to deploy/implement.

tphalp (approved):

PR #43, opened 2024-05-02
Ct/finicity fix

Craiting:
Finicity doesn't have have a way to create connections via an api so you have to create a customer via the api and then generate a widget_url with that customer. So in our app we're using the Oauth flow and webhooks to know when the connection has been made in the other window. The widget will successfully create a connection with Finicity but it's currently not returning any data properly, and Finicity probably doesn't have any VC endpoints to get the data from. This PR is just to get the connection working and write a test similar to the MX OAuth test.

The team should discuss what to do about getting Finicity data working or even if we want to support that in the MVP.

### DIFF
--- a/cypress/e2e/finicity.cy.ts
+++ b/cypress/e2e/finicity.cy.ts
@@ -1,0 +1,15 @@
+describe('Should connect to an Institution through happy path', () => {
+  beforeEach(() => {
+    cy.setAuthCode()
+  })
+
+  it('Connects to Finbank profiles a', () => {
+    const authCode = Cypress.env('authCode')
+
+    cy.visit(`http://localhost:8080/?job_type=agg&auth=${authCode}`)
+    cy.findByPlaceholderText('Search').type('finbank profiles')
+    cy.findByLabelText('Add account with finbank profiles - a').first().click()
+
+    cy.findByRole('link', { name: 'Continue' }).should('exist')
+  })
+})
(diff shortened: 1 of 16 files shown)


server/providers/finicity.ts
```diff
@@ -31,12 +27,12 @@ export class FinicityApi implements ProviderApiClient {
   }
 
   async GetInstitutionById (id: string): Promise<Institution> {
-    const ins = this.apiClient.getInstitution(id)
+    const institution = await this.apiClient.getInstitution(id)
```
wesrisenmay-mx, line 30: At a minimum we should unit test the functions we touched in this file

cypress/e2e/finicity.cy.ts
```diff
@@ -0,0 +1,15 @@
...
+  beforeEach(() => {
+    cy.setAuthCode()
+  })
+
+  it('Connects to Finbank profiles a', () => {
+    const authCode = Cypress.env('authCode')
+
+    cy.visit(`http://localhost:8080/?job_type=agg&auth=${authCode}`)
+    cy.findByPlaceholderText('Search').type('finbank profiles')
+    cy.findByLabelText('Add account with finbank profiles - a').first().click()
```
wesrisenmay-mx, line 11: Do you need the .first()?

wesrisenmay-mx (approved):
Nice work.

cypress/e2e/finicity.cy.ts
Craiting: It turns out that I don't. Removed

wesrisenmay-mx (approved):

PR #46, opened 2024-05-08
Fix issue with missing uuid types, update widget package version, update readme with better monorepo info

tphalp:
Fixes an issue with `uuid` package missing types. This error was being thrown in docker. Docker Compose uses `npm i --omit=dev` so any devDeps aren't installed. Since the `@types/uuid` package was in devDeps, it was missing in the docker build, and was throwing the error.

Add more details about how to update npm packages within the context of the monorepo.

### DIFF
--- a/docker-server.Dockerfile
+++ b/docker-server.Dockerfile
@@ -21,7 +21,6 @@ COPY . ${WRKDR}
 RUN turbo prune --scope=${APP} --docker
 
 FROM base as builder
-RUN npm i -g turbo
 ARG APP
 ARG WRKDR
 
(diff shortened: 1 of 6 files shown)


docker-server.Dockerfile
```diff
@@ -21,7 +21,6 @@ COPY . ${WRKDR}
 RUN turbo prune --scope=${APP} --docker
 
 FROM base as builder
-RUN npm i -g turbo
```
tphalp, line 24: This command wasn't needed....accidental copy/paste.

wesrisenmay-mx (approved):

PR #51, opened 2024-05-16
Wes/deprecate secret exchange use local

wesrisenmay-mx:


Craiting (dismissed):

apps/server/src/config.ts
```diff
@@ -2,52 +2,71 @@ const processEnv = {}
 const envs = { ...process.env, ...(process as Record<string, any>).client_envs }
 
 Object.keys(envs).forEach((k) => {
-  (processEnv as Record<string, any>)[k.toUpperCase()] = envs[k]
+  ;(processEnv as Record<string, any>)[k.toUpperCase()] = envs[k]
```
Craiting, line 5: Is that semi-colon a typo?

Craiting (approved):
Well done

apps/server/src/config.ts
wesrisenmay-mx: Prettier is doing it, and I don't know why

PR #52, opened 2024-05-16
Ct/oauth test provider

Craiting:
Adds tests for Oauth provider

### DIFF
--- a/apps/server/src/providers/index.test.ts
+++ b/apps/server/src/providers/index.test.ts
@@ -4,9 +4,21 @@ import { ProviderApiBase } from './index'
 import { createClient } from '../__mocks__/redis'
 import { ConnectionStatus, OAuthStatus } from '../../src/shared/contract'
 import { READ_MEMBER_STATUS_PATH } from '../test/handlers'
+import { MxApi } from './mx'
 
 const redisMock = createClient()
 const testConnectionId = 'test_connection_id'
+
+const mxApi = new MxApi(
+  {
+    mxProd: {
+      username: 'testUsername',
+      password: 'testPassword'
+    },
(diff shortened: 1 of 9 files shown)


apps/server/src/providers/index.test.ts
```diff
@@ -0,0 +1,114 @@
...
+})
+
+describe('ProviderApiBase', () => {
+  describe('getOauthState', () => {
+    beforeAll(async () => {
+      await providerApiBase.init()
+      providerApiBase.serviceClient = mxApi
+    })
+
+    describe('oauth status PENDING', () => {
```
wesrisenmay-mx, line 41: Typically I use describe for functions, and it for tests cases for those functions. so it("returns a pending oauth state"

apps/server/src/providers/index.test.ts
```diff
@@ -0,0 +1,114 @@
...
+            guid: 'test_connection_id',
+            inbound_member_guid: 'test_connection_id',
+            outbound_member_guid: 'test_connection_id',
+            auth_status: OAuthStatus.PENDING
+          }
+        })
+      })
+    })
+
+    describe('oauth status COMPLETE', () => {
```
wesrisenmay-mx, line 65: Can you create a ticket that we can do later to write the rest of the tests for this file? I just don't want it to get forgotten because we finished some of the tests in here

wesrisenmay-mx (approved):

apps/server/src/providers/index.test.ts
Craiting: Good idea: https://universalconnect.atlassian.net/browse/UCP-133

PR #53, opened 2024-05-21
Ct/institutions local

Craiting:
* Adding a base institution list via the ucw_institution_mapping.json file which is a db export of MX production institution data.
* Using ElasticSearch to index the json file and make it searchable
* Created a basic institution resolver that knows how to use the new institution search


.github/workflows/setup-env.yml
```diff
@@ -30,4 +30,15 @@ jobs:
             echo 'EOF'
           } >> $GITHUB_OUTPUT
           
-          rm ./vars.txt
\ No newline at end of file
+          rm ./vars.txt
+      - name: Configure sysctl limits
```
wesrisenmay-mx, line 34: I don't think this should live here. If this is going to be reused, then we can create another reusable workflow, but it seems like the only place this will be used is for e2e tests

apps/server/src/connect/connectApi.ts
```diff
@@ -29,6 +32,24 @@ function mapInstitution (ins: Institution) {
   })
 }
 
+function mapInstitutionLocal (ins: LocalInstitution): InstitutionResponse {
```
wesrisenmay-mx, line 35: Why don't we modify mapInstitution instead of making mapInstitutionLocal? I think we want the code to behave exactly the same regardless of where the cached data is populated from

apps/server/src/providers/index.ts
```diff
@@ -107,61 +107,37 @@ export class ProviderApiBase {
...
-    }
-    if (query?.length >= 3) {
-      const list = await this.searchApi.institutions(query, this.providers)
-      return list?.institutions?.sort(
-        (a: any, b: any) => a.name.length - b.name.length
-      )
-    }
-    return []
-  }
+  // async search(query: string) {
```
wesrisenmay-mx, line 114: Can we remove this?

apps/server/src/server.js
```diff
@@ -1,37 +1,59 @@
...
+})
+
+app.get('/ping', function (req, res) {
+  res.send('ok')
+})
+
+app.get('/es-index', async (req, res) => {
+  await ElasticsearchClient.deleteIndex()
+  res.send('deleted index')
+  // await ElasticsearchClient.reIndexElasticSearch()
```
wesrisenmay-mx, line 50: Can we remove this comment?

apps/server/src/server.js
```diff
@@ -1,37 +1,59 @@
...
+  info('App initialized successfully')
+}).catch(error => {
+  _error(`Failed to initialized: ${error}`)
+})
+
+app.get('/ping', function (req, res) {
+  res.send('ok')
+})
+
+app.get('/es-index', async (req, res) => {
```
wesrisenmay-mx, line 47: Is this being used?

apps/server/src/server.js
```diff
@@ -1,37 +1,59 @@
...
+  await ElasticsearchClient.deleteIndex()
+  res.send('deleted index')
+  // await ElasticsearchClient.reIndexElasticSearch()
+  // res.send('Indexed institutions')
+})
+
+app.get('/pong/:search_term', async (req, res) => {
```
wesrisenmay-mx, line 54: Is this being used?

apps/server/src/test/testData/institution.ts
```diff
@@ -61,3 +63,60 @@ export const finicityInsitutionData = {
     }
   }
 }
+
+export class LocalInstitutionFactory {
```
wesrisenmay-mx, line 67: Why does this live in a testData file?

Why are we adding "Local" to the name? Don't we expect the institutions to look the same regardless of whether they're in our local default file or coming from the API?

apps/server/src/utils/ElasticSearchClient.ts
```diff
@@ -0,0 +1,114 @@
...
+    try {
+      await client.indices.delete({
+        index: 'institutions'
+      })
+    } catch {
+      console.log('Elasticsearch "institutions" index did not exist')
+    }
+    await client.indices.create({ index: 'institutions' })
+    console.log('Elasticsearch indexing institutions')
+    const dataFilePath = 'ucw_institution_mapping.json'
```
wesrisenmay-mx, line 43: Can we name file what we want it to be named after MVP 1? Like "defaultInstitutions.json"

apps/server/src/utils/ElasticSearchClient.ts
```diff
@@ -0,0 +1,114 @@
...
+    const rawData = readFileSync(dataFilePath)
```
wesrisenmay-mx, line 44: I think we should build this so it is more easily added to when we have an institutions api we are polling.

On initial load it should pull from the json file and load it into the cache. After initialization we should only be looking in the cache.

apps/server/src/utils/ElasticSearchClient.ts
```diff
@@ -0,0 +1,114 @@
...
+    })
+
+    const favoriteInstitutionsResponse: estypes.MgetRequest = await client.mget({
+      docs: esSearch
+    })
+    const institutions = favoriteInstitutionsResponse.docs.map(favoriteInstitution => favoriteInstitution._source as LocalInstitution)
+    return institutions
+  }
+
+  static async deleteIndex () {
```
wesrisenmay-mx, line 103: Can you think of a way that we can unit test this file?

apps/server/src/utils/InstitutionResolver.ts
```diff
@@ -0,0 +1,50 @@
...
+      id: institution.mx.id,
+      url: institution.url,
+      name: institution.name,
+      logo_url: institution.logo,
+      provider: institution.is_test_bank ? 'mx_int' : 'mx'
+    }
+  }
+}
+
+export function availableProviders (institution: LocalInstitution): string[] {
```
wesrisenmay-mx, line 35: I'd call this getAvailableProviders

apps/server/src/utils/InstitutionResolver.ts
```diff
@@ -0,0 +1,50 @@
+import type { InstitutionProvider, LocalInstitution, Provider, ResolvedInstitution } from '../shared/contract'
```
wesrisenmay-mx, line 1: This file needs tests

apps/server/src/utils/InstitutionResolver.ts
```diff
@@ -0,0 +1,50 @@
...
+    }
+    return {
+      id: institutionProvider.id,
+      url: institution.url,
+      name: institution.name,
+      logo_url: institution.logo,
+      provider: provider as Provider
+    }
+  } else {
+    // default to MX for now until further logic is added
```
wesrisenmay-mx, line 24: Do we not have any institutions that mx isn't a provider for currently?

wesrisenmay-mx: I think we should write this as if there are institutions that we don't support in our list. We can make a defaultProvider const in this file that we can fix when we do preferences, but this needs to handle if the defaultProvider isn't in the list of providers.

apps/server/src/utils/InstitutionResolver.ts
```diff
@@ -0,0 +1,50 @@
+import type { InstitutionProvider, LocalInstitution, Provider, ResolvedInstitution } from '../shared/contract'
+import ElasticsearchClient from './elasticSearchClient'
+
+export async function resolveInstitutionProvider (institutionId: string): Promise<ResolvedInstitution> {
```
wesrisenmay-mx, line 4: This function needs some tests

apps/server/src/utils/ElasticSearchClient.ts
tphalp: Yeah, I agree. The filename should probably reflect the date it was generated, or some type of versioning.

.github/workflows/e2e-cypress.yml
```diff
@@ -48,6 +66,11 @@ jobs:
 
           cat ${ENV_FILE_PATH}
 
+      - name: Wait for Elasticsearch to be ready
```
wesrisenmay-mx, line 69: You could add this to cypress' wait-on prop

apps/server/src/connect/connectApi.ts
Craiting: Both methods are needed because one is for mapping institutions which come from elasticsearch results and one maps institutions which come from provider apis and the data formats are different between them.

apps/server/src/utils/ElasticSearchClient.ts
Craiting: Yeah I was thinking the same thing about a timestamp. Changing the filename would be easy.

apps/server/src/utils/ElasticSearchClient.ts
Craiting: Should we save this work for when we have an institutions api to get data from?

apps/server/src/utils/InstitutionResolver.ts
Craiting: All the institutions are from MX's database so MX is a provider for every one. I assumed everything in here would change drastically, I just wanted to create something that did the bare minimum to unblock myself. Now that tests are passing I would prefer to merge this and modify the institutionResolver in a separate MR and ticket for creating the resolver. What do you think?

apps/server/src/utils/InstitutionResolver.ts
Craiting: This function will change a lot but I'll add some tests for it's current state.

Craiting:
There are a couple more things I'd like to refactor a bit. I will mark it as "ready for review" and post in the slack channel when I'm ready for another review.

apps/server/src/utils/ElasticSearchClient.ts
wesrisenmay-mx: I don't think I'm a fan of the timestamp. It adds complexity to loading the file and introduces the potential for people to leave extra copies of their defaults. Everyone should be using versioning, and that will tell them how stale their data is.

apps/server/src/utils/InstitutionResolver.ts
wesrisenmay-mx: I'm working on the resolver right now, so we can leave this super simple

apps/server/src/connect/connectApi.ts
wesrisenmay-mx: The improved naming helps

apps/server/src/utils/ElasticSearchClient.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export const ElasticSearchMock = new Mock()
+
+export const ElasticsearchClient = new Client({
+  node: config.ELASTIC_SEARCH_URL ?? 'http://localhost:9200',
+  ...(process.env.NODE_ENV === 'test' && { Connection: ElasticSearchMock.getConnection() })
+})
+
+export async function initialize (client: Client) {
```
wesrisenmay-mx, line 27: Do we need to pass this client around, or can these functions just call ElasticsearchClient directly

Craiting: I don't think so I was planning on refactoring that this morning

README.md
```diff
@@ -37,6 +37,7 @@ resulting in the error. To fix this, you will need to run `docker logout` from y
 1. Run `cp ./.env.example ./.env`
 1. Run `cp ./apps/server/.env.example ./apps/server/.env`
 1. Follow [Initial Setup](#initial-setup) (below) for setting-up some required environment variables
+1. Install [ElasticSearch] via docker (https://www.elastic.co/guide/en/elasticsearch/reference/current/install-elasticsearch.html) `docker pull elasticsearch:8.13.4`
```
wesrisenmay-mx, line 40: I'm pretty sure I didn't have to do this, but I did have to make sure I had Rancher Desktop installed

apps/server/src/connect/connectApi.test.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  it('finds the institution', async () => {
+    const institution = await connectApi.loadInstitutionByUcpId('UCP-1234')
+    expect(institution).toEqual(expectedInstitutionResponse)
+  })
+})
+
+describe('loadPopularInstitutions', () => {
+  beforeAll(() => {
```
wesrisenmay-mx, line 80: I think you're looking for beforeEach, but because this isn't being reused it should just live inside the it statement

apps/server/src/utils/ElasticSearchClient.test.ts
```diff
@@ -0,0 +1,202 @@
+import { elasticSearchInstitutionData } from '../test/testData/institution'
+import { ElasticSearchMock, getFavoriteInstitutions, getInstitution, initialize, reIndexElasticSearch, search } from '../utils/ElasticSearchClient'
+
+describe('initialize', () => {
+  describe('elastic search already indexed', () => {
+    let indexCreated: boolean = false
+
+    beforeAll(() => {
```
wesrisenmay-mx, line 8: Because there is no reuse here these should live in the it block

apps/server/src/utils/ElasticSearchClient.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    it('does not reindex institutions', async () => {
+      await initialize()
+      expect(indexCreated).toBeFalsy()
+    })
+  })
+
+  describe('elastic search not indexed', () => {
+    let indexCreated: boolean
+
+    beforeAll(() => {
```
wesrisenmay-mx, line 35: Same thing here

apps/server/src/utils/ElasticSearchClient.test.ts
```diff
@@ -0,0 +1,202 @@
...
+      expect(indexCreated).toBeTruthy()
+    })
+  })
+})
+
+describe('reIndexElasticSearch', () => {
+  let indexCreated: boolean
+  let institutionsIndexedCount: number
+
+  beforeAll(() => {
```
wesrisenmay-mx, line 64: Same thing here

apps/server/src/utils/ElasticSearchClient.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    await reIndexElasticSearch()
+    expect(indexCreated).toBeTruthy()
+    expect(institutionsIndexedCount).toBeGreaterThan(4)
+  })
+})
+
+describe('search', () => {
+  beforeEach(() => {
+    ElasticSearchMock.clearAll()
+    ElasticSearchMock.add({
```
wesrisenmay-mx, line 96: I think it makes more sense to put the mock associated with each test in the test instead of in this beforeall

apps/server/src/utils/ElasticSearchClient.test.ts
```diff
@@ -0,0 +1,202 @@
...
+
+  it('makes the expected ES call and returns an empty array', async () => {
+    const results = await search('nothing')
+
+    expect(results).toEqual([])
+  })
+})
+
+describe('getInstitution', () => {
+  beforeEach(() => {
```
wesrisenmay-mx, line 151: Same thing here

apps/server/src/utils/ElasticSearchClient.test.ts
```diff
@@ -0,0 +1,202 @@
...
+  })
+
+  it('makes the expected ES call and gets the expected institution response', async () => {
+    const institutionResponse = await getInstitution('UCP-1234')
+    expect(institutionResponse).toEqual(elasticSearchInstitutionData)
+  })
+})
+
+describe('getFavoriteInstitutions', () => {
+  beforeAll(() => {
```
wesrisenmay-mx, line 171: Same thing here

apps/server/src/utils/ElasticSearchClient.ts
```diff
@@ -0,0 +1,100 @@
...
+import Mock from '@elastic/elasticsearch-mock'
+import { readFileSync } from 'fs'
+import { resolve } from 'path'
+import config from '../config'
+
+import type { CachedInstitution } from 'src/shared/contract'
+
+function getInstitutionFilePath () {
+  if (config.Env === 'test') {
+    console.log('loading test institutions')
```
wesrisenmay-mx, line 12: If we feel like we need to log, then we should use the logger, so we can disable them for our tests. Whenever we run tests right now they are really polluted with log statements, and it makes it difficult to fix things that actually matter.

apps/server/src/utils/institutionResolver.test.ts
```diff
@@ -0,0 +1,126 @@
...
+  })
+})
+
+describe('resolveInstitutionProvider', () => {
+  beforeEach(() => {
+    ElasticSearchMock.clearAll()
+  })
+
+  describe('when MX Test bank', () => {
+    beforeEach(() => {
```
wesrisenmay-mx, line 63: Same thing with these beforeEachs

apps/ui/package.json
```diff
@@ -6,6 +6,7 @@
   "packageManager": "npm@10.5.0",
   "scripts": {
     "dev": "npx cross-env BROWSER=none && vite",
+    "dev:cypress": "npx cross-env BROWSER=none && vite",
```
wesrisenmay-mx, line 9: This could just be npm run dev right?

README.md
Craiting: Oh yeah I think you're right, the run command should pull the image if it doesn't exist. I'll remove this line.

wesrisenmay-mx (approved):

PR #60, opened 2024-06-06
getting search results a little better

Craiting:
Fixing search to not break tests


apps/server/cypress/e2e/fuzzySearch.cy.ts
```diff
@@ -1,16 +1,25 @@
 describe('Should be able to find certain banks with keywords and misspellings', () => {
-  it('Finds expected banks', () => {
+  beforeEach(() => {
```
wesrisenmay-mx, line 2: Abstracting this is fine, but I don't think it adds a lot of value. It requires me to look outside of the tests for a single line of code that's easy to repeat

wesrisenmay-mx (approved):

PR #62, opened 2024-06-06
Use docker volume to save elasticsearch index

Craiting:
This will make it so elasticsearch doesn't have to reinitialize every time the container restarts.


tphalp (approved):

tphalp:
Very nice!

PR #66, opened 2024-06-11
Ct/filter by supported providers

Craiting:
Issue: https://universalconnect.atlassian.net/browse/UCP-157

Filters out institutions which are not supported by a provider in your list of supported providers.


tphalp (approved):
Nice!

apps/server/cypress/e2e/preferences.cy.ts
```diff
@@ -24,4 +24,15 @@ describe('preferences', () => {
...
     cy.findByText('CanDo CU').should('not.exist')
   })
+
+  it('filters out institutions which are not supported by a provider in your list of supported providers.', async () => {
+    cy.visitAgg()
+
+    // "Unsupported Provider" is a test institution included in the institution list solely for this test
+    // it will never show in search as long as the supportedProviders preference works because it doesn't
+    // have any providers configured.
+    cy.findByPlaceholderText('Search').type('Unsupported Provider')
```
wesrisenmay-mx, line 34: I don't love this test. It will continue passing if Unsupported Provider gets removed in the future, or if when the institutions cache is updated Unsupported Provider isn't in there.

Ideally we would test in our regular e2e tests that an institution shows up, and then in a separate e2e test suite we would run a specific test that checks that that same institution doesn't show up with different preferences.

Craiting: I see. My thinking was that it would never be removed, but it's true that we wouldn't know if for some reason it was removed because tests would just keep passing. I'm curious if we're going to need to build out a separate test suite for every variation of preferences.

PR #67, opened 2024-06-12
add test for test bank filtering

Craiting:
https://universalconnect.atlassian.net/browse/UCP-151

If env is prod then test banks will be filtered out


tphalp (approved):
:shipit:

tphalp (approved):

PR #70, opened 2024-06-17
Add another test suite

Craiting:


apps/server/cachedDefaults/testData/alternateTestPreferences.json
```diff
@@ -0,0 +1,63 @@
+{
+  "defaultProvider": "mx",
+
+  "supportedProviders": [
+    "sophtron"
+  ],
+
+  "defaultProviderVolume": {
```
wesrisenmay-mx, line 8: I think ideally these preferences would only have what they need to pass. We could get rid of all the institutionProviderVolumeMap, and all the references to mx

.github/workflows/e2e-cypress-alt.yml
```diff
@@ -0,0 +1,96 @@
...
+          echo FINICITYAPPKEY=${{ secrets.FINICITY_APP_KEY }} >> ${ENV_FILE_PATH}
+
+          cat ${ENV_FILE_PATH}
+
+      - name: Cypress run
+        uses: cypress-io/github-action@v6
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          spec: apps/server/cypress/e2e/suite2/*.cy.ts
```
wesrisenmay-mx, line 86: Instead of using spec we should probably define a second cypress config that points to the tests we want to run with this.

The way it is right now if I do cypress:run locally, then the tests will fail because it's running a test it probably shouldn't be.

wesrisenmay-mx: It seems like we will need more commands for locally running the tests.

cypress:open
cypress:open:alternate

cypress:run
cypress:run:alternate

apps/server/cypress/e2e/suite2/unsupportedProvider.cy.ts
```diff
@@ -0,0 +1,9 @@
+describe('unsupported provider', () => {
```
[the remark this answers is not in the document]
Craiting: Something unexplainable was happening but after trying a bunch of different things it seems like it's working now. It looked like it would find some elements before fetch requests would finish but then a fetch would finish and put the dom back into a loading state, so it wouldn't find MX Bank because it was loading. But now when I test it locally, it seems to work.
image

apps/server/cypress.config.ts
```diff
@@ -4,6 +4,7 @@ export default defineConfig({
   e2e: {
     retries: 1,
     defaultCommandTimeout: 15000,
+    specPattern: 'cypress/e2e/suite2/**/*.{js,jsx,ts,tsx}',
```
wesrisenmay-mx, line 7: You mean suite1 right?

Craiting: Yeah... I don't know how the tests passed with that typo. Something weird is going on.

wesrisenmay-mx (approved):

PR #72, opened 2024-06-21
Ct/filter search by job type

Craiting:
Issue: https://universalconnect.atlassian.net/browse/UCP-156
Adds some filtering by job type including multiple job types. Query considers available providers.

### DIFF
--- a/apps/server/cypress/support/commands.ts
+++ b/apps/server/cypress/support/commands.ts
@@ -32,12 +32,17 @@ Cypress.Commands.add('visitAgg', () => {
   cy.visit('http://localhost:8080/?job_type=agg')
 })
 
-export {}
+Cypress.Commands.add('visitIdentity', () => {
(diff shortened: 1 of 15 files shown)


apps/server/cypress/support/commands.ts
```diff
@@ -32,12 +32,17 @@ Cypress.Commands.add('visitAgg', () => {
   cy.visit('http://localhost:8080/?job_type=agg')
 })
 
-export {}
+Cypress.Commands.add('visitIdentity', () => {
+  cy.visit('http://localhost:8080/?job_type=identify')
```
wesrisenmay-mx, line 36: I know I originally implemented this pattern, but we should configure the baseUrl so we don't have to keep specifying it for each visit: https://docs.cypress.io/api/commands/visit

wesrisenmay-mx (approved):

PR #74, opened 2024-06-26
Wes/sophtron integration tests

wesrisenmay-mx:


apps/server/package.json
```diff
@@ -71,6 +71,7 @@
     "@types/jest": "^29.5.12",
     "@typescript-eslint/eslint-plugin": "^6.20.0",
     "babel-plugin-js-logger": "^1.0.17",
+    "cross-env": "^7.0.3",
```
tphalp, line 74: I feel like this could be a devDep, since it's only being used for tests.

tphalp (approved):
One simple comment about `crossenv`

apps/server/package.json
wesrisenmay-mx: It is a dev dependency, but where it collapsed this file sure made it look like it was a regular dependency.

Craiting (approved):
Excellent work

apps/server/package.json
tphalp: Oh got it. Haha.

PR #76, opened 2024-06-28
Ct/add data endpoints

Craiting:


apps/server/cypress/e2e/suite1/getMxVcData.cy.ts
```diff
@@ -0,0 +1,88 @@
+describe('Create a connection in agg mode and get vc data', () => {
```
Craiting, line 1: This is pretty messy but it is working. We pretty much need to do this same test but for each different job type, and for sophtron. The sophtron vc data structure might be slightly different from MX's, so you might need to put a debugger in at the point where it gets the data to see what the attribute is for `accountGuid`

apps/server/src/server.js
```diff
@@ -3,9 +3,7 @@ import 'dotenv/config'
 import { json, urlencoded } from 'body-parser'
 import express from 'express'
 import config from './config'
-// import http from './infra/http'
 import { wget as _wget, stream } from './infra/http'
-// import { wget as _wget, stream } from './infra/http/real'
 import useConnect from './connect/connectApiExpress'
 import { error as _error, info } from './infra/logger'
 // import useVcs from './incubationVcs/vcsServiceExpress'
```
tphalp, line 9: Any reason to keep this commented import? Line 9.

apps/server/src/services/vcProviders/sophtronVc.test.ts
```diff
@@ -0,0 +1,37 @@
...
+  sophtronVcTranscationsData
+} from '../../test/testData/sophtronVcData'
+import getVC from './sophtronVc'
+
+describe('getVc', () => {
+  const connectionId = 'connectionId'
+  const userId = 'userId'
+  const accountId = 'accountId'
+  const startTime = '1/1/2024'
+  const endTime = '6/1/2024'
```
tphalp, line 13: Is having hard-coded dates a concern? Will these dates work years down the road?

tphalp (approved):
Nice work. Just a few comments.

apps/server/package.json
```diff
@@ -69,6 +71,7 @@
     "@testing-library/cypress": "^10.0.1",
     "@types/crypto-js": "^4.2.2",
     "@types/jest": "^29.5.12",
+    "@types/parse-full-name": "^1.2.5",
```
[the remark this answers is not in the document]
wesrisenmay-mx: Not sure why these were added. Removed them.

apps/server/src/server.js
wesrisenmay-mx: Removed

apps/server/src/services/vcProviders/sophtronVc.test.ts
wesrisenmay-mx: They're parameters being passed to an api that is being mocked, so they could actually be the word junk and it the tests would still pass.

tphalp (approved):
:shipit:

PR #84, opened 2024-07-09
Clean-up in server.js

tphalp:


apps/server/src/server.js
```diff
@@ -1,5 +1,4 @@
 import 'dotenv/config'
-/* eslint-disable @typescript-eslint/no-misused-promises */
```
tphalp, line 2: We can get rid of this `eslint-disable` if we remove the multiple `async` below, and use `void`, instead of `await`.

apps/server/src/server.js
```diff
@@ -40,12 +38,13 @@ app.get('/ping', function (req, res) {
 })
 
 useConnect(app)
-// useVcs(app);
```
tphalp, line 43: Remove unused code

Craiting (approved):
Seems good to me

PR #85, opened 2024-07-10
Consider JobType when resolving an institution

Craiting:
Issue: https://universalconnect.atlassian.net/browse/UCP-163

Resolver checks jobtype to see if provider supports it.

### DIFF
--- a/apps/server/src/services/institutionResolver.ts
+++ b/apps/server/src/services/institutionResolver.ts
@@ -1,5 +1,5 @@
 import { debug } from '../infra/logger'
-import { getInstitution } from '../services/ElasticSearchClient'
+import { getInstitution, JOB_ES_MAPPING } from '../services/ElasticSearchClient'
 import type {
   CachedInstitution,
   InstitutionProvider,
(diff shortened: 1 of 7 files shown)


apps/server/src/services/institutionResolver.ts
```diff
@@ -88,18 +91,61 @@ export async function resolveInstitutionProvider(
...
   }
 
   return providers as Provider[]
 }
+
+function providerSupportsJobType(
+  institutionAttributes: InstitutionProvider | undefined,
+  jobType: JobType
+): boolean {
+  if (institutionAttributes == null) {
```
tphalp, line 120: I feel like a better check here would be: 

`if (!institutionAttributes) {`

Checking for null is different than a variable or parameter that hasn't been passed in, which generally would be `undefined`. 

If you know that the possible values of `institutionAttributes` will sometimes be `null`, then the check you have will work as expected.

tphalp (approved):
Just one comment, but not critical.

apps/server/src/services/institutionResolver.ts
tphalp: Actually, because you are doing a loose comparison (`==`), it will probably work as expected.

wesrisenmay-mx: Agreed on the !

wesrisenmay-mx (approved):

PR #87, opened 2024-07-15
Ct/delete user

Craiting:
Issue: https://universalconnect.atlassian.net/browse/UCP-213

### DIFF
--- a/apps/server/src/connect/connectApiExpress.js
+++ b/apps/server/src/connect/connectApiExpress.js
@@ -19,6 +19,7 @@ import {
   getInstitutionsHandler
 } from './institutionEndpoints'
 import stubs from './instrumentations.js'
+import { userDeleteHandler } from './userEndpoints'
 
 const AGGREGATION_JOB_TYPE = 0
 
@@ -35,10 +36,10 @@ export default function (app) {
     ) {
       return next()
     }
-    req.connectService = new ConnectApi(req)
-    if ((await req.connectService.init()) != null) {
+    req.connectApi = new ConnectApi(req)
+    if ((await req.connectApi.init()) != null) {
       if (!req.context.resolved_user_id) {
-        req.context.resolved_user_id = await req.connectService.ResolveUserId(
+        req.context.resolved_user_id = await req.connectApi.ResolveUserId(
           req.context.user_id
         )
       }
@@ -61,7 +62,7 @@ export default function (app) {
       config.AnalyticsServiceEndpoint !== '' &&
       config.AnalyticsServiceEndpoint != null
     ) {
-      const ret = await req.connectService.analytics(req.path, req.body)
+      const ret = await req.connectApi.analytics(req.path, req.body)
       res.send(ret)
     } else {
       // eslint-disable-next-line @typescript-eslint/no-var-requires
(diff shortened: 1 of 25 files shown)


apps/server/src/connect/connectApiExpress.js
```diff
@@ -218,6 +212,16 @@ export default function (app) {
     })
   })
 
+  app.delete(`${ApiEndpoints.USER}/:userId`, async (req, res) => {
```
wesrisenmay-mx, line 215: New endpoint needs a new section in the OAS(open api spec)

wesrisenmay-mx, line 215: I'd put this handler function somewhere else so we can integration test it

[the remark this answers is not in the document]
wesrisenmay-mx: As well as tests for the error handling

openApiDocumentation.json
```diff
@@ -206,6 +206,29 @@
...
+    "/user/:user_id": {
+      "get": {
+        "parameters": [
+          {
+            "description": "The provider for the connected user",
+            "name": "provider",
+            "in": "query",
+            "required": true,
+            "schema": {
+              "type": "string"
```
wesrisenmay-mx, line 219: Looks like I forgot to do this for the data endpoints, but ideally we would put the enum so they can view the options for provider: https://github.com/Universal-Connect-Project/ucw-app/blob/c0755118ec738a6c85d6be5f40d50e335354d4fd/openApiDocumentation.json#L68

openApiDocumentation.json
```diff
@@ -206,6 +206,29 @@
           }
         }
       }
+    },
+    "/user/:user_id": {
+      "get": {
```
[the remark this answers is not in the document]
Craiting: WHoops

apps/server/src/connect/userEndpoints.ts
```diff
@@ -0,0 +1,35 @@
+import type { Response } from 'express'
+import { getProviderAdapter } from '../adapters'
+
+interface UserDeleteQueryParameters {
+  provider: string
+}
+
+export interface UserDeleteRequest {
+  query: UserDeleteQueryParameters
```
wesrisenmay-mx, line 9: It feels weird to have one of these parameters come from the query and one come from the route params

apps/server/src/connect/userEndpoints.test.ts
```diff
@@ -0,0 +1,86 @@
...
+
+    it('responds with a 400 on error', async () => {
+      const res = {
+        send: jest.fn(),
+        status: jest.fn()
+      } as unknown as Response
+
+      const req: UserDeleteRequest = {
+        query: {
+          provider: 'akoya'
```
wesrisenmay-mx, line 72: We should use a provider that we support

openApiDocumentation.json
```diff
@@ -246,6 +249,30 @@
           }
         }
       }
+    },
+    "/user/:user_id": {
+      "delete": {
+        "parameters": [
```
wesrisenmay-mx, line 255: These parameters need to include all parameters including route params

apps/server/src/connect/userEndpoints.ts
Craiting: This is true thinking about this some more and doing some research I think it might have been a mistake to use query params so heavily with the data endpoints. It seems like most commonly with rest apis, route params are used for params which are required and query params are used for optional params like pagination type things or sort order. 

So the delete endpoint might look like:
```javascript
app.delete(`${ApiEndpoints.USER}/:provider/:userId`, userDeleteHandler)
```

And the data endpoints might change to look like:
```javascript
app.get('/:provider/data/:userId/:connectionId/accounts', accountsDataHandler)
```
For the transactions endpoint I think `start_time` and `end_time` would be query params

Craiting: I think it'd probably be good to refactor these endpoints a bit to follow the common rest patterns. I'll make a ticket to do that for this endpoint and the data endpoints.

apps/server/src/connect/userEndpoints.test.ts
Craiting: Actually using an unsupported provider was intentional to trigger the failure. I fixed the test to make that obvious

wesrisenmay-mx (approved):

PR #95, opened 2024-07-18
Add param validation to delete user endpoint

Craiting:
Issue: https://universalconnect.atlassian.net/browse/UCP-232

Adds some validation to the delete user endpoint

### DIFF
--- a/apps/server/src/connect/userEndpoints.test.ts
+++ b/apps/server/src/connect/userEndpoints.test.ts
@@ -1,5 +1,6 @@
 import type { Response } from 'express'
 import { HttpResponse, http } from 'msw'
+import { Providers } from '../shared/contract'
 import { MX_DELETE_USER_PATH } from '../test/handlers'
 import { listUsersData } from '../test/testData/users'
 import { server } from '../test/testServer'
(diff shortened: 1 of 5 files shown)


apps/server/src/connect/userEndpoints.test.ts
```diff
@@ -10,6 +11,27 @@ const user = listUsersData.users[0]
 
 describe('userEndpoints', () => {
   describe('userDeleteHandler', () => {
+    describe('validation', () => {
+      it('responds with a failure if userId is missing', async () => {
```
wesrisenmay-mx, line 15: Need a test that checks if provider is missing as well

apps/server/src/connect/userEndpoints.ts
```diff
@@ -10,11 +12,31 @@ export interface UserDeleteRequest {
   params: UserDeleteParameters
 }
 
+const createProviderValidator = () =>
```
wesrisenmay-mx, line 15: I think this is at least the 3rd time this has been used. We should put this function in a shared file somewhere

Craiting: Oh yeah I meant to do that before creating a PR. I'll fix this

wesrisenmay-mx (approved):

PR #97, opened 2024-07-22
Wes/data endpoint path parameters

wesrisenmay-mx:


apps/server/src/connect/dataEndpoints.ts
```diff
@@ -4,52 +4,37 @@ import Joi from 'joi'
...
-    res.send(error.details[0].message)
-
-    return
-  }
-
-  const { provider, connection_id, user_id } = req.query
+  const { provider, connectionId, userId } = req.params
 
   const providerAdapter = getProviderAdapter(provider)
-  const providerUserId = await providerAdapter.ResolveUserId(user_id)
```
Craiting, line 49: Since you took out provider validation I think this line will cause a 500 error when the provider isn't valid. 

You could move these methods into the "try" clause and then uncomment [this error](https://github.com/Universal-Connect-Project/ucw-app/blob/57e083c465d0c0db0fe7150e91ca7ff8c52c0588/apps/server/src/adapters/index.ts#L42), then on the response you'd probably want to send the error message.

wesrisenmay-mx: I like that idea

apps/server/src/connect/dataEndpoints.test.ts
```diff
@@ -139,82 +76,17 @@ describe('dataEndpoints', () => {
...
-            }
-          } as IdentityRequest,
-          res
-        )
-
-        expect(res.status).toHaveBeenCalledWith(400)
-        expect(res.send).toHaveBeenCalledWith('"user_id" is required')
-      })
-
-      it('responds with a failure if provider isnt valid', async () => {
```
Craiting, line 183: I think this test would be a good one to keep.  I'm still suspicious that the endpoint will throw a 500 when the provider isn't valid because it's not getting caught. But if you keep this test then we could know for sure.

Craiting (approved):
Nice

PR #107, opened 2024-09-03
Institution polling from server

Craiting:
Adds institution polling to update the institution cache list regularly as the database is updated.


apps/server/src/config.ts
```diff
@@ -14,7 +14,9 @@ const nonSensitiveSharedConfig = {
   ServiceName: 'universal_widget',
   CryptoAlgorithm: 'aes-256-cbc',
   SophtronApiServiceEndpoint: 'https://api.sophtron.com/api',
-  SophtronVCServiceEndpoint: 'https://vc.sophtron.com/api/'
+  SophtronVCServiceEndpoint: 'https://vc.sophtron.com/api/',
+  InstitutionCacheListUrl: 'http://localhost:8088/institutions/cacheList',
```
wesrisenmay-mx, line 18: This will have to differ depending on whether it's local, staging, or production, so it should come from env

apps/server/src/services/ElasticSearchClient.test.ts
```diff
@@ -641,3 +660,121 @@ describe('getRecommendedInstitutions', () => {
...
+
+    const infoSpy = jest.spyOn(logger, 'info').mockImplementation(() => {})
+
+    jest
+      .spyOn(fs, 'readFileSync')
+      .mockReturnValue(JSON.stringify(oldInstitutions))
+
+    await deleteRemovedInstitutions(newInstitutions as CachedInstitution[])
+
+    expect(infoSpy).not.toHaveBeenCalled()
```
wesrisenmay-mx, line 718: checking whether delete is called gives more confidence than a log

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -278,3 +276,45 @@ export async function getRecommendedInstitutions(
...
+
+export async function deleteRemovedInstitutions(
+  newInstitutions: CachedInstitution[]
+) {
+  const newInsIds = newInstitutions.map((ins: CachedInstitution) => ins.ucp_id)
+  const oldInstitutions = JSON.parse(
+    readFileSync(getInstitutionFilePath(), 'utf8')
+  )
+  const oldInsIds = oldInstitutions.map((ins: CachedInstitution) => ins.ucp_id)
+  const deletedInstitutionsIds = oldInsIds.filter(
```
wesrisenmay-mx, line 288: I believe this is a O(n^2) solution, and we're looking at 13,000+ institutions in the array. Worst case scenario this performs 169 million checks.

I'd use array .reduce to create an object with keys as the id and true as the value for everything that's included in newInsIds. You can use that object in this filter function instead of .includes and that brings us down to O(n)

apps/server/src/services/auth0Service.test.ts
```diff
@@ -0,0 +1,94 @@
...
+    jest.spyOn(redisClient, 'get').mockResolvedValue(null)
+    const setSpy = jest.spyOn(redisClient, 'set')
+
+    server.use(
+      http.post(config.Auth0TokenUrl, async ({ request }) => {
+        return HttpResponse.error()
+      })
+    )
+
+    await expect(getAccessToken()).rejects.toThrow(
```
wesrisenmay-mx, line 88: Does this still log if you returned with a status code of 500 instead of HttpResponse.error()?

apps/server/src/services/auth0Service.ts
```diff
@@ -0,0 +1,41 @@
...
+      return null
+    }
+    const responseJson = await response.json()
+    await set(REDIS_AUTH_ACCESS_KEY, responseJson.access_token, {
+      EX: responseJson.expires_in
+    })
+
+    return responseJson.access_token
+  } catch (error) {
+    throw new Error('Could not get UCP access token', error)
```
wesrisenmay-mx, line 39: Is this triggering because response.json fails? I don't think fetch would be throwing an actual error

apps/server/src/services/institutionSyncer.test.ts
```diff
@@ -0,0 +1,135 @@
...
+  beforeEach(() => {
+    jest.useFakeTimers()
+  })
+
+  afterEach(() => {
+    jest.useRealTimers()
+    jest.restoreAllMocks()
+  })
+
+  it.only('should call sync over the given interval', async () => {
```
wesrisenmay-mx, line 31: it.only?

apps/server/src/services/institutionSyncer.test.ts
```diff
@@ -0,0 +1,135 @@
...
+
+    const poller = await setInstitutionSyncSchedule(0.25)
+
+    expect(loadCachedSpy).not.toHaveBeenCalled()
+
+    jest.advanceTimersByTime(15000)
+    await Promise.resolve()
+    expect(loadCachedSpy).toHaveBeenCalledTimes(1)
+
+    await Promise.resolve()
```
wesrisenmay-mx, line 44: What are all these Promise.resolves for?

apps/server/src/services/institutionSyncer.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    jest.useFakeTimers()
+  })
+
+  afterEach(() => {
+    jest.useRealTimers()
+    jest.restoreAllMocks()
+  })
+
+  it.only('should call sync over the given interval', async () => {
+    const loadCachedSpy = jest
```
wesrisenmay-mx, line 32: Generally we want to mock things at as low of a level as possible. In this case I'd use MSW to check that the request is made

apps/server/src/services/institutionSyncer.test.ts
```diff
@@ -0,0 +1,135 @@
...
+    await loadCachedInstitutions()
+
+    expect(infoLogSpy).toHaveBeenCalledWith(
+      'Institution Cache List unchanged. Skipping Update'
+    )
+  })
+
+  it('should call update methods if response is 200', async () => {
+    const infoLogSpy = jest.spyOn(logger, 'info')
+    const deleteInsSpy = jest
```
wesrisenmay-mx, line 81: I would mock something lower level here as well.

apps/server/src/services/institutionSyncer.ts
```diff
@@ -0,0 +1,70 @@
...
+        Authorization: `Bearer ${accessToken}`,
+        'If-None-Match': institutionCacheETag,
+        'Cache-Control': 'public'
+      }
+    })
+    if (response.status === RESPONSE_NOT_MODIFIED) {
+      info('Institution Cache List unchanged. Skipping Update')
+    } else if (response.status === SUCCESS_RESPONSE) {
+      info('Updating institution cache list')
+      await updateElasticSearchDocuments(response)
```
wesrisenmay-mx, line 44: I think we should check that there are institutions in the response before we make changes or cache the file. We could take a widget down if the institutions list came back empty.

apps/server/src/services/institutionSyncer.ts
```diff
@@ -0,0 +1,70 @@
...
+  await deleteRemovedInstitutions(newInstitutions)
+  await updateInstitutions(newInstitutions)
+  await updateInstitutionCacheFile(newInstitutions)
+  await setNoExpiration(
+    INSTITUTION_ETAG_REDIS_KEY,
+    cacheInstitutionResponse.headers.get('etag')
+  )
+}
+
+async function updateInstitutionCacheFile(institutions: CachedInstitution[]) {
```
wesrisenmay-mx, line 68: I'm not sure we should be updating the file here. Maybe we should be putting the list into redis.

If we update the file, then it will change while we're developing locally against staging/local databases

apps/server/src/services/ElasticSearchClient.ts
Craiting: You're right that is terribly inefficient. I did a little research and `Set.has` is an O(1) operation so I can just convert the array to a Set and basically do the same thing but it feels a little simpler to me than using reduce although I think it accomplishes the same thing

apps/server/src/services/institutionSyncer.test.ts
Craiting: Whoops

apps/server/src/services/institutionSyncer.test.ts
Craiting: This was a weird thing I discovered when I was trying to test. I was not getting the results I was expecting when I was doing `jest.advanceTimersByTime` and I think it's because of the nature of the async interval, which won't trigger the function until the previous one is resolved. So I think when I add Promise.resolve(), the test advances the time, and then waits for the async function to resolve. In short I think Promise.resolve() tells the test to slow down for a sec. Having two right after each other is probably not necessary though, I'll remove that.

apps/server/src/services/institutionSyncer.ts
Craiting: Yeah, this is probably not necessary and could cause conflict, I'll remove it.

apps/server/src/config.ts
```diff
@@ -54,7 +58,9 @@ const keysToPullFromEnv = [
   'FinicityAppKeyProd',
   'FinicitySecretProd',
 
-  'ELASTIC_SEARCH_URL'
+  'ELASTIC_SEARCH_URL',
+  'INSTITUTION_POLLING_INTERVAL',
+  'INSTITUTION_CACHE_LIST_URL'
```
wesrisenmay-mx, line 63: This is missing from env.example

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -278,3 +276,51 @@ export async function getRecommendedInstitutions(
...
 }
+
+export async function deleteRemovedInstitutions(
+  newInstitutions: CachedInstitution[]
+) {
+  const newInsIdsSet = new Set(
+    newInstitutions.map((ins: CachedInstitution) => ins.ucp_id)
+  )
+
+  const oldInstitutions = JSON.parse(
```
wesrisenmay-mx, line 287: IMO we should be looking at the cache for oldInstitutions. This list is going to be outdated, and that is going to result in the elastic search documents being out of sync.

Craiting: Hmm I was basically treating the `ucwInstitutionMapping.json` file as a cache, I guess that's why originally I was updating that file each time the widget poller triggered an update.

The elastic search docker container is being initialized from the json file every time you stop and start the server, so the elastic search institutions always start out in sync with that file until polling happens and then elastic search syncs with the institution server. 

We could use redis for cache as well, but it probably will introduce some complexity. It could maybe work if on startup we check redis first and if it contains an institution list then load ElasticSearch via redis but if redis doesn't have the list then we load from the json file. That would make it so editing the institution list manually wouldn't do anything unless you delete the redis cache.

Craiting: If we made it so ElasticSearch data persists between restarts then we might not have to worry about these issues and we could plug the oldInstitutions ucp_ids into redis to be used for checking if any institutions need to be deleted from ElasticSearch

Craiting: I made Redis in a docker container so it will reset on every restart and should stay in sync with Elasticsearch

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -48,24 +49,39 @@ export async function reIndexElasticSearch() {
   } catch {
     info('Elasticsearch "institutions" index did not exist')
   }
-  info('Elasticsearch indexing institutions')
```
wesrisenmay-mx, line 51: The name reIndex makes me think this is something that gets called more than once. Doesn't this only happen once?

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -48,24 +49,39 @@ export async function reIndexElasticSearch() {
   } catch {
```
wesrisenmay-mx, line 49: Line 45 don't we already know there isn't an institutions index, because we checked on line 34?

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -48,24 +49,39 @@ export async function reIndexElasticSearch() {
...
     info('Elasticsearch "institutions" index did not exist')
   }
-  info('Elasticsearch indexing institutions')
-  const dataFilePath = getInstitutionFilePath()
-  const rawData = readFileSync(dataFilePath)
-  const jsonData = JSON.parse(rawData.toString())
+  const institutionData = await getCachedInstitutions()
```
wesrisenmay-mx, line 52: The name getCachedInstitutions is making me think that this is coded wrong, because it's pulling institutions from the file.

apps/server/src/services/institutionSyncer.ts
```diff
@@ -0,0 +1,77 @@
...
+  return setIntervalAsync(
+    async () => {
+      info('Checking for institution list updates')
+      await loadCachedInstitutions()
+    },
+    minutes * 60 * 1000
+  )
+}
+
+export const loadCachedInstitutions = async () => {
```
wesrisenmay-mx, line 27: loadCachedInstitutions is also confusing to me.

updateInstitution? syncInstitutions?

apps/server/src/services/institutionSyncer.ts
```diff
@@ -0,0 +1,77 @@
...
+      logWarning('Unauthorized access to institution cache list')
+    }
+  } catch (error) {
+    logWarning(
+      `Unable to get institution cache list from server: ${error.message}`
+    )
+  }
+}
+
+export async function getCachedInstitutionListFromServer(): Promise<Response | null> {
```
wesrisenmay-mx, line 56: fetchInstitutions?

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -48,24 +49,39 @@ export async function reIndexElasticSearch() {
...
+        document: institution
+      })
+    }
+  )
 
   await Promise.all(indexPromises)
 }
 
+async function getCachedInstitutions(): Promise<CachedInstitution[]> {
+  const response = await getCachedInstitutionListFromServer()
```
wesrisenmay-mx, line 73: We check on the scheduled sync to make sure that there are institutions before we choose to use the response from the server, but we're not checking here

apps/server/src/services/institutionSyncer.ts
```diff
@@ -0,0 +1,77 @@
...
+  try {
+    return await fetch(config.INSTITUTION_CACHE_LIST_URL, {
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+        'If-None-Match': institutionCacheETag,
+        'Cache-Control': 'public'
+      }
+    })
+  } catch {
+    return null
```
wesrisenmay-mx, line 68: I'm not sure if this catch is reachable code. I don't think fetch throws errors, but I could be wrong

wesrisenmay-mx: I know you have unit tests for this. What happens if you try to hit a url where nothing is? Does it hit this catch statement?

Craiting: yeah

apps/server/src/services/ElasticSearchClient.ts
Craiting: yeah I thought that too but didn't want to make this PR any bigger with more refactoring

apps/server/src/services/ElasticSearchClient.ts
Craiting: Yeah, I think originally that reIndex function was meant to be re-used but then probably got refactored and didn't get cleaned up.

apps/server/src/services/ElasticSearchClient.ts
Craiting: It's getting the institutions from either the cached file or the institution server `/cacheList` endpoint. Maybe `getCacheInstitutions ` or `getAvailableInstitutions` idk

wesrisenmay-mx: getInstitutions

apps/server/src/services/ElasticSearchClient.ts
wesrisenmay-mx: I don't think it would change the PR size much to make this stuff better. Adding functionality here and not making it all makes sense feels like adding tech debt

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -34,38 +35,56 @@ export async function initialize() {
...
+  )
 
   await Promise.all(indexPromises)
 }
 
+async function getInstitutions(): Promise<CachedInstitution[]> {
+  const response = await fetchInstitutions()
+  if (response?.ok) {
+    info('Elasticsearch indexing from server list')
+    const newInstitutions = await response.json()
```
wesrisenmay-mx, line 69: Need a try catch for response.json(), and I'd do newInstitutions?.length so it doesn't blow up if it's empty

wesrisenmay-mx (approved):

PR #108, opened 2024-09-04
Remove unused code in preparation to create MX Adapter

tphalp:


.eslintrc.cjs
```diff
@@ -4,7 +4,4 @@ module.exports = {
   ignorePatterns: ["apps/**", "packages/**"],
   extends: ["@ucw-app/eslint-config/base"],
   parser: "@typescript-eslint/parser",
-  parserOptions: {
-    project: true,
-  },
```
tphalp, line 9: My linter suggested I remove this.

.eslintrc.js
```diff
@@ -2,9 +2,6 @@
 /** @type {import("eslint").Linter.Config} */
 module.exports = {
   ignorePatterns: ["apps/**", "packages/**"],
-  extends: ["@repo/eslint-config/library.js"],
+  extends: ["@repo/eslint-config/library.cjs"],
   parser: "@typescript-eslint/parser",
-  parserOptions: {
-    project: true,
-  },
```
tphalp, line 9: My linter suggested I remove this.

.eslintrc.js
```diff
@@ -2,9 +2,6 @@
 /** @type {import("eslint").Linter.Config} */
 module.exports = {
   ignorePatterns: ["apps/**", "packages/**"],
-  extends: ["@repo/eslint-config/library.js"],
+  extends: ["@repo/eslint-config/library.cjs"],
```
tphalp, line 5: Was pointing at incorrect file

.prettierrc.json
```diff
@@ -1,5 +1 @@
-{
-  "trailingComma": "none",
-  "singleQuote": true,
-  "semi": false
```
tphalp, line 4: This was being ignored anyway, so I removed it. This also matches what `ucp-hosted-apps` has, which I think is a cleaner setup. We should align this project with that one in the future.

apps/server/.eslintrc.cjs
```diff
@@ -10,6 +10,7 @@ module.exports = {
     project: './tsconfig.json',
   },
   ignorePatterns: [
+    '.eslintrc.cjs',
     '.eslintrc.js',
     'jest.config.js',
     'jestSetup.ts',
```
tphalp, line 16: We should align on one of these files. (`.eslintrc.cjs` and `.eslintrc.js`--we don't need both). I created a ticket for cleaning up the linting in this project.

apps/server/src/__del_incubationVcs/vcsService.ts
```diff
@@ -59,13 +54,13 @@ export class VcsService extends ProviderAdapterBase {
...
-        `failed creating connection with instituionId : ${institution_id}`
+        `failed creating connection with institutionId : ${institution_id}`
       )
       return {
-        error: `failed creating connection with instituionId : ${institution_id}`
+        error: `failed creating connection with institutionId : ${institution_id}`
       }
     }
-    return { error: 'Unable to find instituion, invalid parameters provided' }
+    return { error: 'Unable to find institution, invalid parameters provided' }
```
tphalp, line 63: Typo

apps/server/src/adapters/mx.test.ts
```diff
@@ -468,13 +469,12 @@ describe('mx provider', () => {
...
-            HttpResponse.json({
-              error: {
-                message: EXTENDED_HISTORY_NOT_SUPPORTED_MSG
-              }
-            })
-          )
+          http.post(EXTEND_HISTORY_PATH, () => {
+            return HttpResponse.json(
+              { error: { message: EXTENDED_HISTORY_NOT_SUPPORTED_MSG } },
+              { status: 400 }
```
tphalp, line 475: Without a non-200 status, the code won't `catch`, which is what is necessary with this update to using `mx-platform-node`.

apps/server/src/adapters/mx.ts
```diff
@@ -173,24 +172,29 @@ export class MxAdapter implements WidgetAdapter {
...
-      ret = await this.apiClient.aggregateMember(request.id, userId)
-    }
-
-    if (ret?.data?.error?.message === EXTENDED_HISTORY_NOT_SUPPORTED_MSG) {
-      ret = await this.apiClient.aggregateMember(request.id, userId)
-    }
 
-    if (ret.data?.error) {
-      return { id: request.id, error_message: ret.data.error.message }
+    try {
```
tphalp, line 176: The `mx-platform-node` library has updated changes, including how it returns errors, if/when they happen. 

Old code returned a 200, but included an `error` object in the body. 

The new library throws an AxiosError, thus the new try/catch pattern.

apps/server/src/adapters/mx.ts
```diff
@@ -319,22 +323,31 @@ export class MxAdapter implements WidgetAdapter {
...
     }
+
     logger.trace(`Creating mx user ${userId}`)
-    const ret = await this.apiClient.createUser({
-      user: { id: userId }
-    })
-    if (ret?.data?.user != null) {
-      return ret.data.user.guid
+
+    try {
```
tphalp, line 340: Same thing as above...new try/catch in order to properly catch the error.

apps/server/src/connect/userEndpoints.test.ts
```diff
@@ -75,7 +75,7 @@ describe('userEndpoints', () => {
 
       await userDeleteHandler(req, res)
 
-      expect(res.send).toHaveBeenCalledWith('')
+      expect(res.send).toHaveBeenCalledWith('User delete failed')
```
tphalp, line 78: For some reason this is different than before. I assume it is because of how errors are handled in the actual mx platform sdk.

apps/server/src/providerApiClients/mx.ts
```diff
@@ -1,11 +1,9 @@
 import providerCredentials from '../providerCredentials'
 import { Configuration, MxPlatformApiFactory } from './mxClient'
-export * from './mxClient/api'
-export * from './mxClient/configuration'
 
 export const BASE_PATH = 'https://api.mx.com'.replace(/\/+$/, '')
 
-export const MxProdApiClient = MxPlatformApiFactory(
+export const MxProdApiClient: any = MxPlatformApiFactory(
```
tphalp, line 6: Linter doesn't like it without a type. It throws a strange error that I've never seen before, but this fixes it. The actual Type is very messy, and unfortunately, the mx-platform-node library doesn't cleanly define a type.

apps/server/src/server.js
```diff
@@ -1,9 +1,9 @@
 import ngrok from '@ngrok/ngrok'
-import { json, urlencoded } from 'body-parser'
```
tphalp, line 2: These are included in express, so we don't need this extra library.

apps/server/src/shared/providers.ts
```diff
@@ -41,7 +41,7 @@ export function getAvailableProviders({
   supportedProviders?: Provider[]
   shouldRequireFullSupport: boolean
 }): Provider[] {
-  const providers = supportedProviders?.filter(
+  return supportedProviders?.filter(
```
tphalp, line 44: Remove redundant variable.

apps/server/src/shared/connect/fireflyApiBridge/index.js
```diff
@@ -17,7 +17,7 @@ const stub = {
   loadUserFeatures: async () => await sendStubData('user_features'),
   createAnalyticsSession: async () => await sendStubData('analytics_sessions'),
   loadJob: async (guid) =>
-    await Promise.resolve({
+    ({
```
tphalp, line 20: Remove redundant Promise

apps/server/src/test/testData/members.ts
```diff
@@ -1,5 +1,3 @@
-import { ConnectionStatus } from '../../shared/contract'
```
tphalp, line 1: Remove unused import

apps/ui/package.json
```diff
@@ -6,6 +6,7 @@
   "packageManager": "npm@10.5.0",
   "scripts": {
     "dev": "npx cross-env BROWSER=none && vite",
+    "dev:debug": "npm run dev",
```
tphalp, line 9: Allows debugging of entire monorepo

package.json
```diff
@@ -6,14 +6,15 @@
...
+    "dev:debug": "turbo dev:debug",
     "dev:e2e": "turbo dev:e2e",
     "lint": "turbo lint",
     "test": "turbo run test",
     "test:watch": "turbo run test:watch",
     "format": "prettier --write \"**/*.{ts,tsx,md}\"",
     "playwright": "playwright test --ui",
     "pre-commit": "lint-staged",
-    "prepare": "husky install",
+    "prepare": "husky",
```
tphalp, line 17: `husky install` is deprecated in favor of `husky`.

turbo.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://turbo.build/schema.json",
+  "$schema": "https://turbo.build/schema.v1.json",
```
tphalp, line 2: This is the preferred way to reference the schema. The versionless url is old.

apps/server/src/shared/providers.ts
```diff
@@ -1,13 +1,13 @@
-import {
+import { JobTypeSupports, MappedJobTypes } from './contract'
+
+import type {
   CachedInstitution,
   InstitutionProvider,
-  JobTypeSupports,
-  MappedJobTypes,
```
tphalp, line 5: Remove unused imports

apps/server/src/providerApiClients/mxClient/index.ts
```diff
@@ -1,17 +1 @@
...
- * NOTE: This class is auto generated by OpenAPI Generator (https://openapi-generator.tech).
- * https://openapi-generator.tech
- * Do not edit the class manually.
- */
-
-
-export * from "./api";
-export * from "./configuration";
-
+export * from 'mx-platform-node'
```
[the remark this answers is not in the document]
tphalp: Yeah, you're right.

.eslintrc.cjs
tphalp: I feel like the way I set up the linting in the project isn't super great. I'm not sure we're using the `packges/eslint-config` the way it's supposed to be used.

tphalp: I found this https://typescript-eslint.io/blog/parser-options-project-true/#introducing-true. I may try using `parserOptions.tsconfigRootDir` to tell it which directory to look in.

apps/server/src/connect/userEndpoints.test.ts
tphalp: What would be a better `it` description?

wesrisenmay-mx: It's not using any specific stuff related to a provider api failing. I'd just call it "responds with a failure if mx deletion api fails"

tphalp: Done

PR #109, opened 2024-09-04
Wes/second test adapter

wesrisenmay-mx:


apps/server/src/shared/contract.ts
```diff
@@ -1,3 +1,4 @@
+import type { TEST_EXAMPLE_A_PROVIDER_STRING } from '../test-adapter'
```
tphalp, line 1: I don't see this type used in this file.

wesrisenmay-mx: Good catches

tphalp (approved):

tphalp (approved):

PR #111, opened 2024-09-12
fixed issue where favorites wouldn't return if it wasn't found in ela…

wesrisenmay-mx:
…stic search


apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -291,6 +291,7 @@ export async function getRecommendedInstitutions(
     })
 
   const institutions = recommendedInstitutionsResponse.docs
+    .filter(({ _source }) => _source)
```
tphalp, line 294: Nice fix.

tphalp (approved):

PR #112, opened 2024-09-12
Fix docker builds and add test to workflow

tphalp:
This PR fixes the docker issues, and adds a test to make sure the docker images will run properly after being built. 

I also added logic to skip the publish step unless the branch being merged to is 'main'. Prior to this PR, this workflow only ran upon merge to main. These changes make this workflow run on every commit, but only publish on merge-to-main.

Resolves UCP-162
Resolves UCP-355
Resolves UCP-363


apps/server/package.json
```diff
@@ -26,8 +26,10 @@
     "@babel/eslint-parser": "^7.18.2",
     "@capacitor-community/http": "^1.4.1",
     "@elastic/elasticsearch": "^8.13.1",
+    "@elastic/elasticsearch-mock": "^2.0.0",
     "@ngrok/ngrok": "^1.2.0",
     "@repo/utils": "*",
+    "@types/crypto-js": "^4.2.2",
```
tphalp, line 32: Types are used in normal code, so needed to be moved here.

apps/server/package.json
```diff
@@ -26,8 +26,10 @@
     "@babel/eslint-parser": "^7.18.2",
     "@capacitor-community/http": "^1.4.1",
     "@elastic/elasticsearch": "^8.13.1",
+    "@elastic/elasticsearch-mock": "^2.0.0",
```
tphalp, line 29: This lib is used in normal code, so needed to be moved here. Docker build skips devDeps

docker-compose.yml
```diff
@@ -1,5 +1,3 @@
-version: '3.8'
```
tphalp, line 1: This is now deprecated.

docker-server.Dockerfile
```diff
@@ -4,8 +4,8 @@
 # the UCW-APP project, which this Dockerfile is part of.
 ARG WRKDR=/opt/app
 
-FROM alpine:3.19.1 as base
-ENV NODE_VERSION 20.12.2
+FROM alpine:3.20.3 as base
```
tphalp, line 7: Update alpine image to resolve some vulnerabilities.

docker-server.Dockerfile
```diff
@@ -43,10 +44,9 @@ RUN npm i -g ts-node  \
 USER nodejs
 
 COPY --from=pruner --chown=nodejs:nodejs ${WRKDR}/out/full/apps/${APP}/ .
+COPY --from=pruner --chown=nodejs:nodejs ${WRKDR}/out/full/packages/utils/ ./packages/utils
 COPY --from=builder --chown=nodejs:nodejs ${WRKDR}/node_modules/ ./node_modules
 
-ENV Env=prod
```
tphalp, line 48: Honor .env file. If we force prod, we will have to have prod MX credentials, which none of us have.

docker-ui.Dockerfile
```diff
@@ -50,8 +51,6 @@ COPY --from=builder --chown=nodejs:nodejs ${WRKDR}/apps/${APP}/vite.config.ts .
 COPY --from=builder --chown=nodejs:nodejs ${WRKDR}/apps/${APP}/dist .
 COPY --from=builder --chown=nodejs:nodejs ${WRKDR}/package.json .
 
-ENV Env=prod
```
tphalp, line 53: Env isn't used in the UI

package.json
```diff
@@ -27,7 +27,7 @@
     "lint-staged": "^15.2.2",
     "prettier": "^3.2.5",
     "prettier-eslint": "^16.3.0",
-    "turbo": "^1.13.3",
+    "turbo": "^2.1.1",
```
tphalp, line 30: Successfully migrated to v2.x of Turborepo

turbo.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema": "https://turbo.build/schema.json",
   "globalDependencies": [],
-  "pipeline": {
+  "tasks": {
```
tphalp, line 4: Changes from v2 migration (turborepo provides a migration script).

wesrisenmay-mx (approved):

wesrisenmay-mx:
We have no way of knowing that this is working without adding it to a workflow

tphalp:
> We have no way of knowing that this is working without adding it to a workflow

I added a new task to do just that.

apps/server/src/server.js
```diff
@@ -28,10 +28,10 @@ const limiter = RateLimit({
 app.use(limiter)
 
 initializeElastic()
-  .then((_) => {
```
tphalp, line 31: unused param

apps/server/src/server.js
```diff
@@ -28,10 +28,10 @@ const limiter = RateLimit({
...
+  .then(() => {
     info('App initialized successfully')
     setInstitutionSyncSchedule(config.INSTITUTION_POLLING_INTERVAL)
-      .then((_) => {
```
tphalp, line 34: unused param

apps/server/src/services/storageClient/redis.ts
```diff
@@ -35,11 +35,14 @@ export const get = async (key: string) => {
   try {
     const ret = await redisClient.get(key)
     return JSON.parse(ret)
-  } catch {}
+  } catch {
+    error('Failed to get value from Redis')
+  }
```
tphalp, line 40: Empty "catch" was throwing linting errors

packages/eslint-config/base.json
```diff
@@ -22,21 +21,5 @@
...
-          "/^react/",
-          "module"
-        ],
-        "alphabetize": {
-          "order": "asc",
-          "ignoreCase": true
-        }
-      }
-    ]
-  }
```
tphalp, line 41: This was causing linting errors, because the order-imports.

PR #122, opened 2024-10-10
Minor fixes, and prep for adapter packages

tphalp:


apps/server/src/infra/context.ts
```diff
@@ -1,43 +1,43 @@
-import type { NextFunction, Request, Response } from "express"
-import config from "../config"
-import { encrypt, decrypt } from "../utils"
```
tphalp, line 3: These two imports weren't used. Also added `type` to other two imports.

tphalp: The rest of the changes to this file are from Prettier.

apps/server/src/services/storageClient/redis.ts
```diff
@@ -1,86 +1,88 @@
...
+import { resolve } from "path";
+import { createClient } from "redis";
+import config from "../../config";
 
-import { debug, error, info } from '../../infra/logger'
-import { PREFERENCES_REDIS_KEY } from './constants'
+import { debug, error, info } from "../../infra/logger";
+import { PREFERENCES_REDIS_KEY } from "./constants";
+
+export * as REDIS_CONSTANTS from "./constants";
```
tphalp, line 9: This export is necessary to expose constant that mx adapter needs.

tphalp: Other changes in this file are from Prettier.

package.json
```diff
@@ -18,24 +18,16 @@
...
-    "@types/node": "^20.12.12",
-    "concurrently": "^8.2.2",
-    "eslint": "^8.57.0",
-    "husky": "^9.0.11",
-    "lint-staged": "^15.2.2",
-    "prettier": "^3.2.5",
-    "prettier-eslint": "^16.3.0",
-    "turbo": "^2.1.1",
-    "typescript": "^5.4.5",
-    "wait-on": "^7.2.0"
```
tphalp, line 33: Some of these devDeps do not need to be here.

packages/utils/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@repo/utils",
   "version": "0.0.1",
-  "main": "./src/index.ts",
+  "main": "index.ts",
```
tphalp, line 4: Fix incorrect path.

turbo.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$schema": "https://turbo.build/schema.v1.json",
+  "$schema": "https://turbo.build/schema.v2.json",
```
tphalp, line 2: Should have been changed in previous PR that upgraded from v1 to v2 or turborepo. Does not cause issues, as far as I could tell, but should match what we are using.

Craiting (approved):

apps/server/src/services/storageClient/redis.ts
wesrisenmay-mx: Why doesn't whatever needs it import it from the constants file?

tphalp: The new mx adapter package needs it. It's being passed in as a dependency in `adapterSetup.ts` in the forked repo.

tphalp: After looking this over, I actually don't need this constant. The entire redis cache instance is being passed in. I'll test it without this export to make sure it works. If so, I'll push up a new PR with this reverted.

PR #125, opened 2024-10-18
Documentation updates for the adapter development process

tphalp:


ADAPTERS.md
```diff
@@ -1,28 +1,40 @@
...
+
+This repo accesses adapter-specific logic in [adapterIndex.ts](./apps/server/src/adapterIndex.ts)
 
-An example adapter lives [here](./apps/server/src/test-adapter/index.ts). Each adapter needs to export an adapterMapObject. These adapterMapObjects are imported and setup by hosters of the UCW in [adapterSetup.ts](./apps/server/src/adapterSetup.ts). If the test adapterMapObject doesn't support a property you need to use for your adapter, then you'll need to add it to the test example adapter to ensure it continues to get support.
+## Monorepo 
 
-This repo accesses adapter specific logic via [adapterIndex.ts](./apps/server/src/adapterIndex.ts)
+Because this repo is a monorepo, there are some caveats to consider with regard to adapter creation. See [the MONOREPO readme](MONOREPO.md#creating-an-aggregator-adapter-package-within-a-monorepo) for more details.
```
emilysallstrom, line 40: This feels like it's making the Monorepo page much more complicated. Since creating adapters is so rarely done, I think it might make sense to keep everything related to it all together, rather than introducing complexity for a fintech engineer (I'm thinking through the different users who will be using this, and this is for the aggregator engineer, and the other page is more geared toward the fintech engineer).

The items on the Monorepo page seem almost like sub-steps under "Build your adapter" within this page. I'm thinking it would make sense to add more detail to each of the steps here. So instead of "Build your adapter", put the detail from the Monorepo packaging. "Add a test institution" could have an example of what the test institution would look like.

This might be a bigger ask, so we could also create a separate ticket for it.

tphalp: Great feedback. So, keep the monorepo readme unchanged, and just add more details to the Adapters readme?

emilysallstrom: Exactly :+1:

tphalp: Okay, @emilysallstrom I made some changes, including adding info about test institutions. Give it a look when you have a minute. Thank you!

ADAPTERS.md
```diff
@@ -1,28 +1,98 @@
...
-1. Add a test institution to the [default institution list](./apps/server/cachedDefaults/ucwInstitutionsMapping.json) with aggregator support defined for your institution.
+1. Add a test institution to the [default institution list](./apps/server/cachedDefaults/ucwInstitutionsMapping.json) with aggregator support defined for your institution (See [below](#test-institutions))
 1. Build your adapter
 1. Import your adapter in [adapterSetup.ts](./apps/server/src/adapterSetup.ts) on your forked version of this repo
 1. Write unit, integration, and e2e tests
 1. Bundle and publish your adapter to npm
-1. When your adapter package is ready for production use, then you'll need to gain access to the UCP institution list to update the list of institutions you support(This functionality is in progress by the UCP team)
+1. When your adapter package is ready for production use, then you'll need to gain access to the UCP institution list to update the list of institutions you support (This functionality is in progress by the UCP team)
+
+We are open to contributions to make this process easier.
```
wesrisenmay-mx, line 32: What do you mean by this?

ADAPTERS.md
```diff
@@ -1,28 +1,98 @@
...
+
+Below you'll find some tips on how to successfully run the UCW with a compiled package.
+
+We will use [the MX Adapter](https://github.com/Universal-Connect-Project/ucw-adapter-mx/tree/main/packages/mx-adapter) as a reference.
+
+#### Development tips
+
+When you create an adapter package, you'll need to install it in your monorepo workspace using the following command:
+
+```bash
```
wesrisenmay-mx, line 83: Someone creating an adapter package doesn't need to do this right? Isn't this only for those hosting?

wesrisenmay-mx:
Nice work.

ADAPTERS.md
```diff
@@ -1,28 +1,98 @@
...
 1. Fork this repository
 1. Create a folder in the [packages](./packages) folder for your adapter
-1. Add a test institution to the [default institution list](./apps/server/cachedDefaults/ucwInstitutionsMapping.json) with aggregator support defined for your institution.
+1. Add a test institution to the [default institution list](./apps/server/cachedDefaults/ucwInstitutionsMapping.json) with aggregator support defined for your institution (See [below](#test-institutions))
 1. Build your adapter
 1. Import your adapter in [adapterSetup.ts](./apps/server/src/adapterSetup.ts) on your forked version of this repo
 1. Write unit, integration, and e2e tests
 1. Bundle and publish your adapter to npm
-1. When your adapter package is ready for production use, then you'll need to gain access to the UCP institution list to update the list of institutions you support(This functionality is in progress by the UCP team)
+1. When your adapter package is ready for production use, then you'll need to gain access to the UCP institution list to update the list of institutions you support (This functionality is in progress by the UCP team)
```
emilysallstrom, line 30: We should create a ticket to update this once we have the institution list live

emilysallstrom (dismissed):
Looks good, thanks Tyson!

ADAPTERS.md
tphalp: I'll remove it. I'm not sure why I put that there.

ADAPTERS.md
tphalp: That is intended for anyone creating a new adapter. In order to test it with `app/server`, they need to add it to the server package file. I'll add some more details to make that more clear.

ADAPTERS.md
tphalp: Done. https://universalconnect.atlassian.net/browse/UCP-416

ADAPTERS.md
```diff
@@ -1,28 +1,101 @@
...
+#### Development tips
+
+When developing an adapter package, you'll need to install it in your monorepo workspace in order to test it. You can do this using the following command:
+
+```bash
+npm install <your_package_name> --workspace apps/server
+```
+_Package name is the `name` property in your adapter package's `package.json`_
+
+Once you run the above command, change the version of the package to a start with `*`. In the context of npm, this means install the latest version of the package. This is also a convention of Turborepo, and is what is recommended when developing package locally.
```
wesrisenmay-mx, line 84: Are you sure this works? The turborepo internal package documentation says to use @repo/ for internal packages. This feels like it's running off a previously installed version of the package that was added during the npm install command.

tphalp: @wesrisenmay-mx `@repo/` is just a convention that they picked because there aren't any packages on npm with the `@repo` scope. @repo doesn't really have any special meaning in the turborepo context. 

This screenshot is taken from the turborepo docs: 
image

I've tested this quite extensively. If the output directory exists (in this case, `dist`), and the proper `exports` are set up in the package file, then the project will use the local assets, even if the package exists in the npmjs repository.

tphalp: @wesrisenmay-mx I added the additional step.

wesrisenmay-mx (approved):

PR #126, opened 2024-10-18
Sync changes from the MX Adapter fork

tphalp:


.github/workflows/docker-test.yml
```diff
@@ -54,6 +54,18 @@ jobs:
...
+          DOCKER_IMAGE_SERVER=$(npm pkg get version --prefix ./apps/server | xargs)
+          echo DOCKER_IMAGE_SERVER="${DOCKER_IMAGE_SERVER}" >> ./.env
+          echo DOCKER_IMAGE_SERVER="${DOCKER_IMAGE_SERVER}" >> $GITHUB_ENV
+
+      - name: 'Get UI version'
+        run: |
+          DOCKER_IMAGE_UI=$(npm pkg get version --prefix ./apps/ui | xargs)
+          echo DOCKER_IMAGE_UI="${DOCKER_IMAGE_UI}" >> ./.env
+          echo DOCKER_IMAGE_UI="${DOCKER_IMAGE_UI}" >> $GITHUB_ENV
+
```
tphalp, line 68: Code that was should be here. While it doesn't cause issues if it's missing, it is a better mirror of the actual docker build process if this code is here.

.github/workflows/unit-tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     needs: [setup-env]
 
     steps:
-      - name: Checkout under $GITHUB_WORKSPACE
+      - name: Checkout repository
```
tphalp, line 17: The variable in the old version wasn't doing anything. It just showed up in the logs as the variable, and not the value you would expect. I don't think the variable is useful anyway, if we were to make this work as was originally expected.

wesrisenmay-mx (approved):

PR #139, opened 2024-11-15
Some small updates

tphalp:
This PR is a collection of necessary changes that were identified while working on the Sophtron Adapter package.


apps/server/src/test-adapter/constants.ts
```diff
@@ -10,6 +10,10 @@ export const TEST_EXAMPLE_C_LABEL_TEXT = "TestExampleC Label";
 
 export const testAggregatorMemberGuid = "testAggregatorMemberGuid";
 
+export const testDataValidatorStartTimeError =
```
wesrisenmay-mx, line 13: These are only used once and are test specific right? I'd probably hardcode them in the tests instead

tphalp: I'm actually using them in the test-adapter as return values for the DataRequestValidation.transactions return value, so I'm using them in two places each.

apps/server/src/connect/dataEndpoints.test.ts
```diff
@@ -233,29 +261,6 @@ describe("dataEndpoints", () => {
         expect(res.status).toHaveBeenCalledWith(400);
       });
 
-      it("doesn't respond with a 400 if it's TestAdapterA and there is no start or end time", async () => {
```
tphalp, line 236: Moved down with the other validation tests.

apps/server/src/connect/dataEndpoints.test.ts
```diff
@@ -397,6 +402,77 @@ describe("dataEndpoints", () => {
...
+        } as unknown as any;
+
+        await createTransactionsDataHandler(true)(req, res);
+
+        expect(res.send).toHaveBeenCalledWith({
+          jwt: transactionsResponse,
+        });
+      });
+
+      it("does NOT respond with a 400 if it's TestAdapterA and start_time is undefined", async () => {
```
tphalp, line 430: This is the one moved from above.

apps/server/src/connect/dataEndpoints.test.ts
```diff
@@ -26,9 +34,29 @@ const vcAccountsDataHandler = createAccountsDataHandler(true);
...
+  aggregator: TEST_EXAMPLE_B_AGGREGATOR_STRING,
+};
+let connectTestB: ConnectApi;
+
 describe("dataEndpoints", () => {
-  beforeEach(() => {
+  beforeEach(async () => {
     jest.restoreAllMocks();
+
+    connectTestA = new ConnectApi({
```
wesrisenmay-mx, line 51: Why are these necessary? I see where they're passed in down below, but I don't think they're actually being used.

tphalp: If I don't do this, then the `req.connectApi` property isn't populated, and the code doesn't work.

apps/server/src/connect/dataEndpoints.test.ts
```diff
@@ -397,6 +402,77 @@ describe("dataEndpoints", () => {
         expect(res.send).toHaveBeenCalledWith("Something went wrong");
         expect(res.status).toHaveBeenCalledWith(400);
       });
+
+      it("returns transaction data, if it's TestAdapterB and passes the transaction validator", async () => {
```
wesrisenmay-mx, line 406: These it statements are hard to parse. When looking at the code it should be easy for me to look through the tests and see if there's coverage for everything. We don't need to mention which aggregator it's for except for sophtron. the it statements should focus on the code that's being tested instead of the aggregator that's being tested. 

Examples: "succeeds if there is a custom validator and it passes" "succeeds if there's no custom validator" "fails if a custom validator fails" "fails if it's sophtron and it's missing a startTime" "succeeds if it's sophtron and it has a startTime" etc

apps/server/src/connect/dataEndpoints.test.ts
```diff
@@ -27,7 +30,7 @@ const vcIdentityDataHandler = createIdentityDataHandler(true);
 const vcTransactionsDataHandler = createTransactionsDataHandler(true);
 
 describe("dataEndpoints", () => {
-  beforeEach(() => {
+  beforeEach(async () => {
```
wesrisenmay-mx, line 33: No longer need this async

apps/server/cypress/e2e/suite1/testExample.cy.ts
```diff
@@ -38,6 +40,96 @@ const makeABConnection = async (jobType) => {
...
+      method: "GET",
+      url: `/api${url}`,
+      failOnStatusCode: false,
+    })
+    .then((dataResponse) => {
+      expect(dataResponse.status).to.equal(400);
+    });
+};
+
+export const generateValidatorTests = ({ makeAConnection }) => {
```
wesrisenmay-mx, line 91: The generateValidatorTests function feels like a premature abstraction. It's only used once. I'd just put the it statement directly in the describe

packages/utils-dev-dependency/cypress/generateDataTests.ts
```diff
@@ -80,7 +80,7 @@ const verifyTransactions = ({
   shouldTestVcEndpoint,
   userId,
 }) => {
-  const url = `/data/aggregator/${aggregator}/user/${userId}/account/${accountId}/transactions${aggregator === "sophtron" ? "?start_time=2021/1/1&end_time=2024/12/31" : ""}`;
+  const url = `/data/aggregator/${aggregator}/user/${userId}/account/${accountId}/transactions${aggregator === "testExampleB" || aggregator === "sophtron" ? "?start_time=2021/1/1&end_time=2024/12/31" : ""}`;
```
wesrisenmay-mx, line 83: We need this not care about which aggregator it is. Should we pass in an optional query string parameter?

apps/server/cypress/e2e/suite1/testExample.cy.ts
```diff
@@ -46,5 +96,46 @@ describe("testExampleA and B aggregators", () => {
   generateDataTests({
     makeAConnection: makeABConnection,
     shouldTestVcEndpoint: true,
+    queryString: "?start_time=2021/1/1",
+  });
+
+  const jobType = JobTypes.VERIFICATION;
```
wesrisenmay-mx, line 102: I don't think this variable needs to exist

apps/server/cypress/e2e/suite1/testExample.cy.ts
```diff
@@ -38,6 +40,54 @@ const makeABConnection = async (jobType) => {
...
+    return accountId;
+  });
+};
+
+const verifyTransactionsValidatorSuccess = ({
+  accountId,
+  aggregator,
+  userId,
+}) => {
+  const url = `/data/aggregator/${aggregator}/user/${userId}/account/${accountId}/transactions${aggregator === TEST_EXAMPLE_B_AGGREGATOR_STRING ? "?start_time=2021/1/1" : ""}`;
```
wesrisenmay-mx, line 60: no reason for the aggregator === TEST_EXAMPLE_B_AGGREGATOR_STRING right?

packages/utils-dev-dependency/cypress/generateDataTests.ts
```diff
@@ -79,8 +79,16 @@ const verifyTransactions = ({
...
+  queryString = "",
+}: {
+  accountId: string;
+  aggregator: string;
+  shouldTestVcEndpoint: boolean;
+  userId: string;
+  queryString?: string;
 }) => {
-  const url = `/data/aggregator/${aggregator}/user/${userId}/account/${accountId}/transactions${aggregator === "sophtron" ? "?start_time=2021/1/1&end_time=2024/12/31" : ""}`;
+  cy.log("querystring", queryString);
```
wesrisenmay-mx, line 90: no need to log this right?

apps/server/cypress/e2e/suite1/testExample.cy.ts
tphalp: Yep. Fixed.

apps/server/cypress/e2e/suite2/sophtron.cy.ts
```diff
@@ -63,5 +63,9 @@ describe("Sophtron aggregator", () => {
     expectConnectionSuccess();
   });
 
-  generateDataTests({ makeAConnection, shouldTestVcEndpoint: true });
+  generateDataTests({
+    makeAConnection,
+    shouldTestVcEndpoint: true,
+    queryString: "?start_time=2021/1/1&end_time=2022/1/1",
```
wesrisenmay-mx, line 69: This variable name makes it unclear where this is going to be used. Maybe transactionsQueryString?

packages/utils-dev-dependency/cypress/generateDataTests.ts
tphalp: Yep, removed.

apps/server/cypress/e2e/suite2/sophtron.cy.ts
tphalp: Renaming done

apps/server/cypress/e2e/suite1/testExample.cy.ts
```diff
@@ -38,6 +40,55 @@ const makeABConnection = async (jobType) => {
...
+
+    return accountId;
+  });
+};
+
+const verifyTransactionsValidatorSuccess = ({
+  accountId,
+  aggregator,
+  userId,
+  queryString = "",
```
wesrisenmay-mx: Why do we need an aggregator or queryString variable if it's always the same?

tphalp: Good point. Removed

tphalp: Removing aggregator....

tphalp: Done

wesrisenmay-mx (approved):

emilysallstrom (approved):

PR #143, opened 2024-11-21
Remove sophtron and mx code/tests from apps/server

tphalp:


apps/server/src/test/testData/institution.ts
```diff
@@ -4,32 +4,6 @@ import {
...
-};
-
-export const sophtronUserInstitutionAccountsData = [
-  {
-    AccountName: "AccountName",
-    AccountNumber: "AccountNumber",
-    AccountID: "AccountID",
-  },
-];
-
```
tphalp, line 32: These were all unused

apps/server/src/utils/index.ts
```diff
@@ -1,56 +1,6 @@
-import { createCipheriv, createDecipheriv } from "crypto";
```
tphalp, line 1: More unused methods/vars in this file

apps/server/src/adapterSetup.ts
```diff
@@ -1,25 +1,28 @@
...
-    widgetAdapter: new SophtronAdapter(),
-  },
-};
+const sophtronAdapterMapObject: Record<string, AdapterMap> =
+  getSophtronAdapterMapObject({
+    logClient: logger,
+    aggregatorCredentials: {
+      clientId: config.SophtronApiUserId,
+      secret: config.SophtronApiUserSecret,
+      endpoint: config.SophtronApiServiceEndpoint,
```
wesrisenmay-mx, line 17: should endpoint and vcEndpoint just be hardcoded into the adapter?

tphalp: Good catch.

tphalp: We should do the same for the MX adapter.

wesrisenmay-mx:
It's nice to see all this red

wesrisenmay-mx (approved):

PR #146, opened 2024-11-26
Wes/serve UI with express

wesrisenmay-mx:


docker-server.Dockerfile
```diff
@@ -2,52 +2,16 @@
...
+# Copy app source
+COPY . .
+ 
+RUN npm install
+RUN npm run build
 
 EXPOSE ${PORT}
 
-CMD ["ts-node", "./src/server.js"]
+CMD ["npm", "run", "start:server"]
```
tphalp, line 17: You're losing a lot of security in the docker file by removing the code that created a limited user, and instead just running everything as root.

Also, moving from `alpine` to `node` image will make the final docker image much larger. The original Dockerfile that was created by sophtron resulted in an image that was over 1GB is size.  After moving to a slimmer image base, and using different build contexts, the image was around 180MB.

docker-compose.yml
```diff
@@ -18,60 +16,15 @@ services:
...
           platforms:
             - linux/arm64
             - linux/amd64
-    image: universalconnectfoundation/ucw-app-server:v${DOCKER_IMAGE_SERVER}
     networks:
       - ucw
     ports:
       - "${SERVER_PORT}:${SERVER_PORT}"
-    env_file:
-      - ./apps/server/.env
```
tphalp, line 34: If you remove this, how are you accessing the end-user's env file?

wesrisenmay-mx: The env file gets copied over with the rest of the code and loaded through dotenv

tphalp:
https://github.com/Universal-Connect-Project/ucw-app/blob/wes/serveUIWithExpress/README.md?plain=1#L50 should be updated with the new `.../widget?...` path change.

docker-server.Dockerfile
```diff
@@ -2,52 +2,25 @@
...
-
-FROM base AS builder
-ARG APP
-ARG WRKDR
-
-WORKDIR ${WRKDR}
+# Copy app source
+COPY . .
+ 
+RUN npm ci
```
tphalp, line 16: if you do `nom ci --omit=dev` it could trim the size down as well, since we shouldn't need the devDeps for prod.

tphalp (approved):

PR #152, opened 2024-12-04
Wes/toggle data endpoints

wesrisenmay-mx:


Craiting (approved):

apps/server/src/config.ts
```diff
@@ -22,45 +12,39 @@ const nonSensitiveSharedConfig = {
...
 
-  "HOSTURL",
-  "WebhookHostUrl",
-  "LogLevel",
+  "HOST_URL",
+  "WEBHOOK_HOST_URL",
+  "LOG_LEVEL",
 
   "UCPClientId",
   "UCPClientSecret",
```
tphalp, line 20: I can't remember--are we using these anymore?

wesrisenmay-mx: They're used for polling the institution service, but there is no institution service yet

PR #162, opened 2024-12-10
Ct/deploy heroku

Craiting:


apps/server/src/services/storageClient/redis.ts
```diff
@@ -8,6 +8,12 @@ import { PREFERENCES_REDIS_KEY } from "./constants";
 
 const redisClient = createClient({
   url: config.REDIS_SERVER,
+  ...(config.REDIS_ENABLE_TLS && {
+    socket: {
+      tls: true,
+      rejectUnauthorized: false,
+    },
+  }),
```
Craiting, line 16: This config makes Redis work in heroku, but it breaks things locally and in the github workflows, that's why I added an optional env variable which I could just set in Heroku to get it to work.

Procfile
```diff
@@ -0,0 +1 @@
+web: npm run copyDemoPreferences && npm start --workspace=apps/server
```
wesrisenmay-mx, line 1: The widget urls probably won't work like this. The UI needs to be built for the server to work off it's assets

apps/server/package.json
```diff
@@ -9,13 +9,14 @@
...
     "cypress:run:prod": "cypress run --config-file cypress.config.prod.ts",
-    "elasticsearch": "docker run --rm --name elasticsearch_container -p 9200:9200 -p 9300:9300 -e \"discovery.type=single-node\" -e \"xpack.security.enabled=false\" elasticsearch:8.13.4",
+    "elasticsearch": "docker run --rm --name elasticsearch_container -p 9200:9200 -p 9300:9300 -e \"discovery.type=single-node\" -e \"xpack.security.enabled=false\" docker.elastic.co/elasticsearch/elasticsearch:7.10.2",
     "dev": "concurrently --kill-others \"npm run elasticsearch\" \"npm run server\" \"npm run redisserver\"",
     "dev:debug": "concurrently --kill-others \"npm run elasticsearch\" \"npm run server:debug\" \"npm run redisserver\"",
     "start:e2e": "nodemon",
     "redisserver": "docker run --rm --name ucw_redis -p 6379:6379 redis:7.2-alpine",
     "server": "wait-on http://localhost:9200 && nodemon",
     "server:debug": "wait-on http://localhost:9200 && node --inspect -r ts-node/register ./src/server.js",
+    "start": "npx ts-node src/server.js",
```
wesrisenmay-mx, line 19: Should we be using --transpileOnly?

apps/server/src/server.js
```diff
@@ -84,8 +84,9 @@ app.get("*", (req, res) => {
   res.sendStatus(404);
 });
 
-app.listen(config.PORT, () => {
-  const message = `Server is running on port ${config.PORT}, ENV: ${config.ENV}, LOG_LEVEL: ${config.LOG_LEVEL}`;
+const port = process.env.PORT || config.PORT
```
wesrisenmay-mx, line 87: Why do we need this line? config.PORT is getting it's value from process.env.PORT

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -255,7 +279,7 @@ function buildMustNotQuery(hiddenInstitutions: string[]): any[] {
     },
   });
 
-  if (!["test", "dev"].includes(config.ENV)) {
+  if (!["test", "dev", "staging"].includes(config.ENV)) {
```
wesrisenmay-mx, line 282: Why don't we just make this prod?

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -267,12 +291,12 @@ function buildMustNotQuery(hiddenInstitutions: string[]): any[] {
...
 
 export async function getInstitution(id: string): Promise<CachedInstitution> {
-  const institutionResponse = await ElasticsearchClient.get({
+  const { body } = await ElasticsearchClient.get({
     id,
     index: "institutions",
   });
 
-  return institutionResponse._source as CachedInstitution;
+  return body._source as CachedInstitution;
```
wesrisenmay-mx, line 299: Is this change because of openSource?

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -342,31 +370,62 @@ export async function deleteRemovedInstitutions(
...
+      for (const ucpId of deletedInstitutionsIds) {
+        await ElasticsearchClient.delete({
+          index: "institutions",
+          id: ucpId,
+        });
+        console.info(`Deleted institution ${ucpId}`);
+      }
+    } else {
+      await Promise.all(
+        deletedInstitutionsIds.map(async (ucpId: string) => {
```
wesrisenmay-mx, line 385: Looks like you could share this function and use deletedInstitutionsIds.forEach for one a Promise.all for the other

wesrisenmay-mx: Same thing for update and index

wesrisenmay-mx: Ah actually never mind. Can't await with forEach. Should probably create a function for the duplicated code though.

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -342,31 +370,62 @@ export async function deleteRemovedInstitutions(
...
-  const deletePromises = deletedInstitutionsIds.map(async (ucpId: string) => {
-    return await ElasticsearchClient.delete({
-      index: "institutions",
-      id: ucpId,
-    });
-  });
 
   try {
-    await Promise.all(deletePromises);
+    if (config.ELASTIC_SEARCH_SINGLE_THREAD) {
```
wesrisenmay-mx, line 375: Need tests for the new config.ELASTIC_SEARCH_SINGLE_THREAD cases

Procfile
Craiting: This command is what's run after heroku finishes the build phase. The UI gets built during the build phase which comes from the package.json "build" script.

apps/server/package.json
Craiting: Oh yeah good idea

apps/server/src/server.js
Craiting: Oh yeah that's true this is not necessary

apps/server/src/services/ElasticSearchClient.ts
Craiting: Yeah it's the difference between Elasticsearch version 8 and 7.10

apps/server/src/services/ElasticSearchClient.test.ts
```diff
@@ -268,44 +268,51 @@ describe("indexElasticSearch", () => {
...
-          elasticSearchInstitutionData,
-        ]);
-      }),
-    );
+  describe.each([
+    { singleThread: true, description: "with single-threading" },
+    { singleThread: false, description: "without single-threading" },
+  ])("Index Elasticsearch $description", ({ singleThread }) => {
+    it("indexes institutions retrieved from the institution server", async () => {
+      config.ELASTIC_SEARCH_SINGLE_THREAD = singleThread;
```
wesrisenmay-mx, line 276: I would mock the config instead of modifying it directly. This feels like it could leak into other tests. We use getConfig in a few other places to make it mockable

apps/server/src/services/ElasticSearchClient.test.ts
```diff
@@ -646,41 +658,47 @@ describe("getRecommendedInstitutions", () => {
...
-    await overwriteSet(
-      INSTITUTION_CURRENT_LIST_IDS,
-      oldInstitutions.map((ins) => ins.id),
-    );
+  describe.each([
+    { singleThread: true, description: "with single-threading" },
+    { singleThread: false, description: "without single-threading" },
+  ])("Delete institutions $description", ({ singleThread }) => {
+    it("should delete institutions that are no longer in the new list", async () => {
+      config.ELASTIC_SEARCH_SINGLE_THREAD = singleThread;
```
wesrisenmay-mx, line 666: Same thing here

PR #176, opened 2025-01-31
Akoya

sophtron-jishun:
Added mikomo bank, 
Added oauth handling in contract


playwright.config.ts
```diff
@@ -0,0 +1,15 @@
+import { PlaywrightTestConfig, devices } from '@playwright/test';
```
wesrisenmay-mx, line 1: We've been putting test config inside of whatever app/package it relates to, so this should live in the akoya adapter in the akoya folder

packages/utils/contract.ts
```diff
@@ -33,7 +33,7 @@ export enum JobTypes {
 export type AdapterMap = {
   dataAdapter?: Function;
   vcAdapter?: Function;
-  createWidgetAdapter: () => WidgetAdapter;
+  createWidgetAdapter: (sessionId?: string | undefined) => WidgetAdapter;
```
wesrisenmay-mx, line 36: I don't see anywhere that a sessionId gets passed in to createWidgetAdapter.

We generally prefer to pass function parameters in as objects for extensibility, so if we are passing this in we'd prefer { sessionId }

apps/server/src/test-adapter/adapter.ts
```diff
@@ -251,4 +251,14 @@ export class TestAdapter implements WidgetAdapter {
   ): Promise<string> {
     return userId;
   }
+
+  async HandleOauthResponse(request: any): Promise<Connection>{
```
wesrisenmay-mx, line 255: Needs a unit test

apps/server/src/config.ts
```diff
@@ -33,6 +33,11 @@ const keysToPullFromEnv = [
   "FINICITY_APP_KEYProd",
   "FINICITY_SECRETProd",
 
+  "AKOYA_CLIENT_ID",
```
wesrisenmay-mx, line 36: After we import the akoya package and officially have support for these, then we'll need to add them to the environment readme as well as .env.example in the server

apps/server/src/connect/oauthEndpoints.ts
```diff
@@ -0,0 +1,88 @@
...
+import { createAggregatorWidgetAdapter } from "../adapterIndex";
+import { get, set } from "../services/storageClient/redis";
+import { ConnectionStatus } from "../shared/contract";
+import * as path from 'path'
+import * as fs from 'fs'
+import config from "../config";
+import * as logger from "../infra/logger";
+
+//GET '/oauth/:aggregator/redirect_from/'
+export const oauthRedirectHandler =
```
wesrisenmay-mx, line 11: Need unit tests for both these functions

apps/server/src/connect/oauthEndpoints.ts
```diff
@@ -0,0 +1,88 @@
...
+      const queries = {
+        status: ret?.status === ConnectionStatus.CONNECTED ? 'success': 'error',
+        app_url,
+        redirect: ret?.oauth_referral_source?.toLowerCase() === 'browser' ? `false`: 'true',
+        error_reason: ret?.error,
+        member_guid: ret?.id,
+      };
+      
+      const oauthParams =  new RegExp(Object.keys(queries).map(r => `\\$${r}`).join('|'), 'g');
+      function mapOauthParams(queries: any, res: Response, html: string){
```
wesrisenmay-mx, line 46: This function feels like an unnecessary abstraction. It would be less complicated to just put line 47 in place of line 64

apps/server/src/infra/http/oauth/error.html
```diff
@@ -0,0 +1,58 @@
...
+    <link rel="stylesheet" href="%PUBLIC_URL%/index.css" />
+    <script type="text/javascript" src="/oauth/oauth.js" ></script>
+  </head>
+  <body>
+    <div id="container">
+      <svg xmlns="http://www.w3.org/2000/svg" width="32" height="32" viewBox="0 0 32 32" fill="none">
+        <path fill-rule="evenodd" clip-rule="evenodd" d="M16 0C24.8 0 32 7.2 32 16C32 24.8 24.8 32 16 32C7.2 32 0 24.8 0 16C0 7.2 7.2 0 16 0ZM16 22C14.8954 22 14 22.8954 14 24C14 25.1046 14.8954 26 16 26C17.1046 26 18 25.1046 18 24C18 22.8954 17.1046 22 16 22ZM16.04 6C15.2447 6 14.6 6.72754 14.6 7.625V17.375C14.6 18.2725 15.2447 19 16.04 19C16.8353 19 17.48 18.2725 17.48 17.375V7.625C17.48 6.72754 16.8353 6 16.04 6Z" fill="#49505A"/>
+      </svg>
+      <h2 class="title">Something went wrong</h2>
+      <p class="subtitle">We've notified support and we're looking into the issue. Please close this. </p>
```
wesrisenmay-mx, line 21: This line of text doesn't feel true. I don't know that we're doing any tracking of these errors

apps/server/src/infra/http/oauth/oauth.js
```diff
@@ -0,0 +1,39 @@
...
+  var closeButton = document.getElementById('oauth-close-window')
+  if (closeButton) {
+    closeButton.addEventListener('click', function() {
+      window.location = app_url
+    })
+  }
+}
+
+
+// def build_ui_scheme_redirect(ui_scheme, member_guid, error_reason = nil)
```
wesrisenmay-mx, line 25: Need tests for this file + removal of comment

apps/server/src/infra/http/oauth/error.html
```diff
@@ -0,0 +1,58 @@
...
+      <button id="oauth-close-window">Go back to app</button>
+    </div>
+    <script type="text/javascript">
+      
+      function handle_oauth(redirect, app_url, post_message, member_guid, error_reason){
```
wesrisenmay-mx, line 26: Can we move this function into it's own file, so we can unit test it?

wesrisenmay-mx:
We need an end to end test for the new functionality. We've been adding support for all features to the testExample adapter. So we'll need an end to end test for the testExample adapter that uses the new webhook stuff

playwright.config.ts
sophtron-jishun: They playwright tests you added were in packages but the test task in the workflow were initiated from the root folder, besides when there are multiple test framework so it makes things easier  having this config in the root file to seek for consistent file pattern.

apps/server/src/connect/oauthEndpoints.ts
```diff
@@ -0,0 +1,88 @@
...
+import config from "../config";
+import * as logger from "../infra/logger";
+
+//GET '/oauth/:aggregator/redirect_from/'
+export const oauthRedirectHandler = 
+  async (req: Request, res: Response) => {
+    const { userId, aggregator } = req.params;
+    try {
+      const aggregatorAdapter = createAggregatorWidgetAdapter(aggregator);
+      const oauth_res = await aggregatorAdapter.HandleOauthResponse({...req.query, ...req.params, ...req.body})
```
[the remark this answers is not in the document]
sophtron-jishun: this is minimal and exact enough information that may come from an oauth redirect call back. not necessary to pass the req obj in. this makes the handler logic clearer.

apps/server/src/infra/http/oauth/error.html
sophtron-jishun: this is what's carried from mx connect code.  I think we should get some help from some UX design team about this later

apps/server/src/infra/http/oauth/error.html
sophtron-jishun: these were carried from mx connect code too, when I did the test a while ago there were issues (can't remember exactly, such as leading to the window being unable to close itself )  when trying to remove this code into a js file. but I wasn't sure about the exact reasons.  I'm interested in knowing why they had to implement this inline.

apps/server/src/infra/http/oauth/success.html
```diff
@@ -0,0 +1,65 @@
...
+          </clipPath>
+        </defs>
+      </svg>
+      <h2 class="title">Almost there</h2>
+      <p class="subtitle">Please close this and return to your app to continue.</p>
+      <button id="oauth-close-window">Go back to app</button>
+    </div>
+    <script type="text/javascript">
+      
+      function handle_oauth(redirect, app_url, post_message, member_guid, error_reason){
```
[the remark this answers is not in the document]
sophtron-jishun: same as above

apps/server/src/connect/oauthEndpoints.ts
sophtron-jishun: I think this is necessary as it selects different html files to trigger different display and behavior in the client app. although, I understand your point of seeing this redundant, same as the ones below, this behaviour is what I extracted from the original mx connect source code. I believe they did it this way for a reason,  it's better to keep it as is and seek for more context so we can revisit it later.

playwright.config.ts
wesrisenmay-mx: The only reason the workflows are in the root folder is because that's what we're forced to do by github. We want test configuration to live in the package that it's used for. 

MX adapter example: https://github.com/Universal-Connect-Project/ucw-adapter-mx/blob/main/packages/mx-adapter/playwright.config.ts

apps/server/src/connect/oauthEndpoints.ts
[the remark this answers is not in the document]
wesrisenmay-mx: Spreading the query, params, and body into one object makes it difficult to follow where things are coming from. It also introduces potential issues of the body overriding things of the same name in query or params.

This is the akoya adapter HandleOauthReponse code:

`// eslint-disable-next-line @typescript-eslint/no-explicit-any
  async HandleOauthResponse(request: any): Promise<Connection> {
    // eslint-disable-next-line @typescript-eslint/naming-convention
    const { state: request_id, code } = request;
    this.logger.info(`Received akoya oauth redirect response ${request_id}`);
    const connection = await this.cacheClient.get(request_id);
    if (!connection) {
      return null;
    }
    if (code) {
      connection.status = ConnectionStatus.CONNECTED;
      connection.guid = connection.institution_code;
      connection.id = connection.institution_code;
      connection.user_id = code;
      connection.request_id = request_id;
    }
    // console.log(connection)
    await this.cacheClient.set(request_id, connection);

    return connection;
  }`
  
When spreading all the objects into one I can't tell where the request_id or code are coming from, because it could be coming from query, params, or body.

I can think of a few options that would be less confusing

1. Pass the whole request in
2. create a new object that has less information, but doesn't obfuscate where things are coming from { body, query, params}
3. Pass only the information that's used by aggregators into the function

apps/server/src/infra/http/oauth/error.html
wesrisenmay-mx: For now just remove the subtitle

apps/server/src/infra/http/oauth/error.html
wesrisenmay-mx: This is what ChatGPT says about how to do it:

https://chatgpt.com/share/67a0d7c5-6d88-800f-bea5-d24959c6db23

apps/server/src/connect/oauthEndpoints.ts
```diff
@@ -0,0 +1,88 @@
...
+          'utf8', 
+          (err, content) => {
+            if(err){
+              reject(err);
+            }
+            resolve(content);
+          }
+        );
+      }) 
+      mapOauthParams(queries, res, html);
```
wesrisenmay-mx, line 64:
```suggestion
      res.send(html.replaceAll(oauthParams, q => queries[q.substring(1)] || ''))
```

playwright.config.ts
sophtron-jishun: it doesn't hurt anything having a basic config at root level but only makes package development easier

apps/server/src/connect/oauthEndpoints.ts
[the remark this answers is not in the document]
sophtron-jishun: it doesn't matter, it's only to make the interface abstracted clearer and more compatible, for one particular aggregator, it can only come from one way,  which is mostly in the `query`, and whatever other way they pass information with, there won't be conflicts either. 
this way gives precise information to adapter implementation instead of exposing the whole request

apps/server/src/infra/http/oauth/error.html
sophtron-jishun: I don't want to remove the predesigned code and spend time to add it back later,  I think mx should have a design team that made this decision already.

apps/server/src/infra/http/oauth/error.html
sophtron-jishun: as I mentioned I suggest asking mx internal team why they did it this way. not asking a bot actually

playwright.config.ts
wesrisenmay-mx: I disagree

apps/server/src/connect/oauthEndpoints.ts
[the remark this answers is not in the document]
wesrisenmay-mx: Option 2 exposes less information, but makes it more clear where things are coming from

apps/server/src/infra/http/oauth/error.html
wesrisenmay-mx: Are you talking about UX or code design? Abstracting a function out of a file doesn't change the UX, and code design is up to engineers

emilysallstrom: We shouldn't assume that the MX code in the original connect widget is using best practices. This is rendered by our backend Ruby service, Firefly, which is probably why it's inline. We should use best practices here whenever possible, even if it diverts from the MX connect widget.

wesrisenmay-mx: Our strategy is to only commit code that we know we're using and that we can actually test. If we can't test the code, then we leave it out until we can test the code.

apps/server/src/adapters/index.ts
```diff
@@ -64,6 +64,7 @@ export class AggregatorAdapterBase {
       if (this.context?.aggregator) {
         this.aggregatorAdapter = createAggregatorWidgetAdapter(
           this.context?.aggregator as Aggregator,
+          token //use the auth token as sessionId,
```
wesrisenmay-mx, line 67: If you look on line 60 you'll see that token is just "fakeTokenThatWeNeedToRemove"

apps/server/src/adapterIndex.ts
```diff
@@ -5,12 +5,13 @@ import { adapterMap } from "./adapterSetup";
 
 export function createAggregatorWidgetAdapter(
   aggregator: Aggregator,
+  sessionId?: string | undefined
```
wesrisenmay-mx, line 8: This should follow our general pattern of using object parameters instead of lists of parameters

apps/server/package.json
```diff
@@ -23,7 +23,7 @@
     "lint": "eslint . --ext .js,.ts",
     "lint:fix": "eslint . --fix --ext .js,.ts",
     "coverage": "npm test -- --verbose --coverage --watchAll",
-    "test": "jest --no-coverage",
+    "test": "NODE_OPTIONS=\"$NODE_OPTIONS --experimental-vm-modules\" jest --no-coverage",
```
wesrisenmay-mx, line 26: Why do we need this?

apps/server/src/server.js
```diff
@@ -82,6 +83,11 @@ app.get("*", (req, res) => {
   res.sendStatus(404);
 });
 
+if(!config.PORT){
+  console.log(config)
```
wesrisenmay-mx, line 87: Config is going to have secrets in it. We shouldn't log it

apps/server/src/test-adapter/adapter.test.ts
```diff
@@ -342,4 +341,28 @@ describe("TestAdapter", () => {
       );
     });
   });
+  describe("HandleOauthResponse", () => {
+    it("responds success from HandleOauthResponse", async () => {
```
wesrisenmay-mx, line 345: We should be testing all the branching logic of the function

apps/server/src/connect/connectApiExpress.js
```diff
@@ -178,4 +182,12 @@ export default function (app) {
       members: ret,
     });
   });
+  
+  app.all('/webhook/:aggregator/*', webhookHandler)
```
wesrisenmay-mx, line 186: We need e2e tests for these endpoints

apps/server/src/connect/oauthEndpoints.ts
```diff
@@ -0,0 +1,84 @@
...
+
+//ALL '/webhook/:aggregator/*'
+export const webhookHandler = 
+  async (req: Request, res: Response) => {
+    const { aggregator } = req.params;
+
+    try {
+      const aggregatorAdapter = createAggregatorWidgetAdapter(aggregator);
+      logger.info(`received web hook at: ${req.path}`, req.query)
+      //console.log(req.body)
```
wesrisenmay-mx, line 76: Unnecessary comment

apps/server/src/connect/oauthEndpoints.ts
```diff
@@ -0,0 +1,84 @@
...
+import * as logger from "../infra/logger";
+
+//GET '/oauth/:aggregator/redirect_from/'
+export const oauthRedirectHandler = 
+  async (req: Request, res: Response) => {
+    const { userId, aggregator } = req.params;
+    try {
+      const aggregatorAdapter = createAggregatorWidgetAdapter(aggregator);
+      const oauth_res = await aggregatorAdapter.HandleOauthResponse({query: req.query, params: req.params, body: req.body})
+      const ret = {
```
wesrisenmay-mx, line 17: This ret object is pretty confusing. The name "ret" makes it seem like this is going to be returned somewhere or was returned from a request, but neither of those appears correct.

It also has an aggregator property added to it that doesn't appear to be used anywhere.

The ret object makes it harder to follow where things are coming from farther down in the function.

Can we grab parameters directly from oauth_res and context instead?

apps/server/src/connect/oauthEndpoints.ts
```diff
@@ -0,0 +1,84 @@
...
+        ...oauth_res,
+        aggregator
+      } as any
+      if(ret.id){
```
wesrisenmay-mx, line 21: There's quite a bit of branching logic in this function that hasn't been unit tested

apps/server/package.json
wesrisenmay-mx: bump

apps/server/playwright/oauth.playwright.ts
```diff
@@ -0,0 +1,52 @@
...
+  await page.getByPlaceholder("Search").fill("TestExampleFailedOauth Bank");
+
+  await page.getByLabel("Add account with TestExampleFailedOauth Bank").click();
+  
+  const popupPromise = page.waitForEvent("popup");
+  await page.getByRole("link", { name: "Continue" }).click();
+
+  const authorizeTab = await popupPromise;
+  const url = await authorizeTab.url()
+  console.log('popup => ' + url);
```
wesrisenmay-mx, line 21: Unnecessary log

apps/server/playwright/oauth.playwright.ts
```diff
@@ -0,0 +1,52 @@
...
+  await page.getByPlaceholder("Search").fill("TestExampleOauth Bank"); //TestExampleOauth Bank
+
+  await page.getByLabel("Add account with TestExampleOauth Bank").click();
+  
+  const popupPromise = page.waitForEvent("popup");
+  await page.getByRole("link", { name: "Continue" }).click();
+
+  const authorizeTab = await popupPromise;
+  const url = await authorizeTab.url()
+  console.log('popup => ' + url);
```
wesrisenmay-mx, line 46: Unnecessary log

apps/server/playwright/oauth.playwright.ts
```diff
@@ -0,0 +1,52 @@
+import { expect, test } from "@playwright/test";
+
+test("displays error page with failed oAuth", async ({ page }, testInfo) => {
```
wesrisenmay-mx, line 3: Nice tests

apps/server/src/adapters/index.ts
```diff
@@ -57,13 +57,15 @@ export class AggregatorAdapterBase {
   }
 
   async init() {
-    const token = "fakeTokenThatWeNeedToRemove";
-
-    this.analyticsClient = new AnalyticsClient(token);
+    const authToken = this.context?.auth?.token || 'default_token';
```
wesrisenmay-mx, line 60: I believe the auth object is a remnant of the authentication system that was being used on AWS. Given our current documentation this object will be empty. I also have it removed in the branch that I'm using to use the new connect widget frontend: https://github.com/Universal-Connect-Project/ucw-app/blob/connectWidgetReplacement/apps/server/src/connect/instrumentationEndpoints.ts

I'm not sure what would be a good thing to use a sessionId. Maybe we generate a uuid in the instrumentation handler that stays until they refresh?

apps/server/src/connect/connectApiExpress.js
wesrisenmay-mx: Bump

wesrisenmay-mx: I have some examples of endpoint tests here: https://github.com/Universal-Connect-Project/ucw-app/tree/connectWidgetReplacement/apps/server/cypress/e2e/suite1/endpoints

apps/server/src/test-adapter/adapter.ts
```diff
@@ -43,6 +43,22 @@ export class TestAdapter implements WidgetAdapter {
   DataRequestValidators: Record<string, (req: any) => string | undefined> = {};
 
   async GetInstitutionById(id: string): Promise<Institution> {
+    if(id.toLowerCase().indexOf('oauth') >= 0){
```
wesrisenmay-mx, line 46: Nice solution

apps/server/playwright/oauth.playwright.ts
sophtron-jishun: I can remove it if that makes you happy, but it's designed test output

apps/server/package.json
sophtron-jishun: the `prettier` for formatting html strings requires this option to get around some errors

wesrisenmay-mx (approved):

PR #179, opened 2025-02-10
Using combo jobs for favorites, instrumentation, and the widget url

wesrisenmay-mx:


apps/server/openApiInternal.json
```diff
@@ -56,6 +47,18 @@
...
                   },
+                  "jobTypes": {
+                    "type": "array",
+                    "items": {
+                      "type": "string",
+                      "enum": [
+                        "account_verification",
+                        "identity_verification",
+                        "transactions",
+                        "transaction_history"
```
Craiting, line 58: Do the values in this enum need to be snake case? I'm noticing a weird mix of snake and camel case.

Craiting (approved):

apps/server/openApiInternal.json
wesrisenmay-mx: I created a ticket to audit this before we go live with it. These are copy pasted from the connect widget, but we could translate them into camelCase. I don't want to do that until I get a full use case working though

https://universalconnect.atlassian.net/jira/software/projects/UCP/boards/1/timeline?selectedIssue=UCP-516

PR #186, opened 2025-02-25
Wes/sync

wesrisenmay-mx:


apps/server/src/adapters/index.ts
```diff
@@ -171,14 +170,12 @@ export class AggregatorAdapterBase {
     if (ret.auth_status === OAuthStatus.ERROR) {
       ret.error_reason = connection.status;
     }
-    return { oauth_state: ret };
+    return ret;
   }
 
   async getOauthStates(memberGuid: string) {
```
Craiting, line 176: It looks like this is missing a test

apps/server/src/connect/oauthEndpoints.test.ts
```diff
@@ -0,0 +1,189 @@
...
+import { TEST_EXAMPLE_A_AGGREGATOR_STRING } from "../test-adapter";
+import { oauthSuccessResponse } from "./testData/oauth";
+import { format } from "prettier";
+
+const context = {
+  aggregator: TEST_EXAMPLE_A_AGGREGATOR_STRING,
+};
+let connect: ConnectApi;
+
+describe("oauthtHandler", () => {
```
Craiting, line 14:
```suggestion
describe("oauthHandler", () => {
```

apps/server/src/connect/oauthEndpoints.test.ts
```diff
@@ -0,0 +1,189 @@
...
+    } as unknown as any;
+
+    await webhookHandler(req, res);
+
+    expect(res.send).toHaveBeenCalledWith({
+      id: "request_id",
+      request_id: "request_id",
+      guid: "test_guid",
+      user_id: "test_code",
+      status: 6,
```
Craiting, line 149: Do we have a Status object anywhere we can use here? It'd be more readable if it were something like STATUS.PENDING instead of `6`

apps/server/src/connect/oauthEndpoints.test.ts
```diff
@@ -0,0 +1,190 @@
...
+    connect = new ConnectApi({
+      context,
+    });
+    await connect.init();
+  });
+
+  it("responds success from oauthRedirectHandler", async () => {
+    jest.mock("../services/storageClient/redis", () =>
+      jest.requireActual("../__mocks__/redis"),
+    );
```
Craiting, line 26: I don't think this is necessary, I think Redis gets mocked automatically by existing in the __mocks__ directory

apps/server/src/connect/oauthEndpoints.test.ts
```diff
@@ -0,0 +1,190 @@
...
+      context,
+    });
+
+    await connect.init();
+  });
+
+  it("responds success from webhookHandler", async () => {
+    jest.mock("../services/storageClient/redis", () =>
+      jest.requireActual("../__mocks__/redis"),
+    );
```
Craiting, line 113: this too

Craiting (approved):
I don't think redis needs to be explicitly mocked so those lines I commented on could be removed. Otherwise looks good. I'll approve though because I might be wrong about the redis mock.

apps/server/src/connect/oauthEndpoints.test.ts
wesrisenmay-mx: These were from the sync, but I fixed them. Good catch.

PR #191, opened 2025-03-01
allow raw jobtype input, enrich member status update message info

sophtron-jishun:


wesrisenmay-mx:
This PR isn't following our [testing ADR](https://github.com/Universal-Connect-Project/architecture-decision-record/blob/main/architectureDecisionRecords/0002Testing.md), and the unit tests are failing.

apps/server/src/connect/connectApi.ts
```diff
@@ -51,13 +51,15 @@ function mapConnection(connection: Connection): Member {
     institution_guid: connection.institution_code,
     guid: connection.id,
     connection_status: connection.status ?? ConnectionStatus.CREATED, // ?
+    raw_status: connection.raw_status,
```
wesrisenmay-mx, line 54: If you're adding new functionality in a PR, then I'm going to need a description of why it's necessary and how it's used. I can't do a proper review without understanding why the change is even necessary.

With our large time zone difference and long delays in being able to answer questions it would be best if before you submitted a PR for review you ensured that it followed all the ADRs and has enough context for the reviewer to understand it.

We've found a lot of code in ucw-app that ended up not even being used, so we're trying to only commit code that is absolutely necessary, documented, and well tested.

PR #212, opened 2025-04-02
add akoya adapter

sophtron-jishun:


.github/workflows/akoya-adapter.yml
```diff
@@ -0,0 +1,69 @@
...
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: lts/Iron
+      - name: Install dependencies
+        run: npm ci
+
+      - run: npm run copyTestPreferences
```
wesrisenmay-mx, line 45: Shouldn't this be copyAkoyaTestPreferences?

.github/workflows/akoya-adapter.yml
```diff
@@ -0,0 +1,69 @@
...
+
+          # This fails if using https, so we have to override it
+          echo HOST_URL=http://localhost:8080 >> ${ENV_FILE_PATH}
+
+          cat ${ENV_FILE_PATH}
+
+      - name: Install Playwright Browsers
+        run: npx playwright install --with-deps
+      - name: Run Playwright tests
+        run: npm run playwright:test --workspace=apps/server
```
wesrisenmay-mx, line 63: This should be workspace=packages/akoya-adapter right?

.github/workflows/akoya-adapter.yml
```diff
@@ -0,0 +1,69 @@
+name: Playwright Tests
```
wesrisenmay-mx, line 1: Can we put something in this name to make it clear these are the akoya tests?

packages/akoya-adapter/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@repo/akoya-adapter",
+  "version": "0.0.1-alpha",
+  "description": "Akoya Adapter for the Universal Connect Widget",
+  "main": "src/index.ts",
+  "scripts": {
+    "cypress:open": "cypress open --config-file cypress.config.ts",
```
wesrisenmay-mx, line 7: The Akoya adapter doesn't use. cypress right? So we don't need any of these cypress commands or the cypress dependencies?

packages/akoya-adapter/playwright/akoya.playwright.ts
```diff
@@ -0,0 +1,60 @@
+import { expect, test } from "@playwright/test";
```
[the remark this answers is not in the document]
sophtron-jishun: dataendpoints test was added in the akoya adapter test packages/akoya-adapter/playwright/akoya.playwright.ts line 70 and you approved it

however i didn't copy it over because those endpoints are not enabled by config. from the existing code it looks the moved-over adapters( mx , sophtron )  both don't have tests that hits the data endpoints,  thought you have it planed otherwise

packages/akoya-adapter/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "axios": "^1.7.7",
+    "parse-address": "^1.1.2",
+    "parse-full-name": "^1.2.6",
+    "uuid": "^11.0.3"
+  },
+  "devDependencies": {
+    "@playwright/test": "^1.49.1",
+    "@testing-library/cypress": "^10.0.2",
+    "@typescript-eslint/eslint-plugin": "^6.20.0",
+    "cypress": "^13.15.0",
```
wesrisenmay-mx, line 26: We don't need this right?

packages/akoya-adapter/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "dotenv": "^16.4.5",
+    "eslint": "^8.57.1",
+    "eslint-config-standard-with-typescript": "^43.0.1",
+    "eslint-plugin-cypress": "^2.15.1",
```
wesrisenmay-mx, line 30: We don't need this right?

packages/akoya-adapter/playwright/akoya.playwright.ts
```diff
@@ -0,0 +1,130 @@
...
+  const connectedPromise = new Promise((resolve, reject) => {
+    const timer = setTimeout(() => reject('timed out'), 12000);
+    page.on("console", async (msg : any) => {
+      const obj = (await msg.args()[0].jsonValue())?.message
+      if(obj?.type === 'connect/memberConnected'){
+        const { user_guid, member_guid, aggregator } = obj.metadata;
+        const idToken = await apiClient.getIdToken(user_guid);
+        const user = encodeURIComponent(JSON.stringify(idToken))
+        let accountId;
+        let url = `http://localhost:8080/api/data/aggregator/${aggregator}/user/${user}/connection/${member_guid}/accounts`
```
wesrisenmay-mx, line 80: The userId for the data endpoint is expected to be the same userId that was passed in with the widget url that was used to create a connection

sophtron-jishun: akoya doesn't have concept of user, and have no data parameter to consume another user_id (or whatever id) 
this is how akoya should be handled in the uniformed interface

btw, the userId in data endpoints is neither the user_id passed in in the widget url. that's resolved_user_id by aggregator

wesrisenmay-mx: My concern is that this breaks the pattern established by the documentation and the other 2 adapters.

The docs we've put out state that you are supposed to pass the same userId that you passed in with your widget, and that you should be able to get data with that. If Akoya doesn't work the same way, then our users have to write custom code just for Akoya to get their data back.

sophtron-jishun: this is the way how akoya works,   
agregators are different, 

we are doing widget, and the widget is doing connection making, it's never a good idea putting data endpioint in the widget server, as I mentioned.  they target entirely different audience.   widget is supposed to make the connection and provide what clients needs to enable client handle, aggregators have all kinds of different apis for clients to use, 
you are only wrapping up a couple and thinks that the end of story, you are compromising the reality with your preference

packages/akoya-adapter/src/apiClient.ts
```diff
@@ -0,0 +1,153 @@
...
+    this.envConfig = envConfig;
+    this.client_redirect_url = `${this.envConfig.HostUrl}/oauth/${this.apiConfig.aggregator}/redirect_from`;
+    this.authParams = {
+      client_id: apiConfig.clientId,
+      client_secret: apiConfig.secret,
+    };
+  }
+
+  getOauthUrl(institutionId: string, state: string) {
+    return `${this.apiConfig.basePath}/auth?connector=${institutionId}&client_id=${this.apiConfig.clientId}&redirect_uri=${this.client_redirect_url}&state=${state}&response_type=code&scope=openid email profile offline_access`;
```
Craiting, line 59: Is `email` necessary in this scope? When I was testing locally, it didn't work until I removed `email`.

Craiting:
We decided to remove the data adapter stuff because it doesn't fit in the currently defined pattern. So we're calling Akoya support only "partial support". I branched off this branch to remove things and refactored a few things. The code is now merged into main so we can close this PR. Thanks

PR #213, opened 2025-04-08
Wes/refresh with ucp

wesrisenmay-mx:


apps/ui/src/api/connectWidgetApiService.ts
```diff
@@ -32,7 +36,7 @@ const connectWidgetApiService = {
   },
   loadInstitutionByGuid: async (guid: string) => {
     return configuredAxios
-      .get(`/institutions/${guid}`)
+      .get(`/institutions/${institutionId || guid}`)
```
wesrisenmay-mx, line 39: Need a test for this

apps/server/src/institutions/institutionResolver.ts
```diff
@@ -35,72 +36,93 @@ const getAggregatorByVolume = (
...
+  aggregatorOverride?: Aggregator;
+  jobTypes: ComboJobTypes[];
+  ucpInstitutionId: string;
+}): Promise<ResolvedInstitution> {
+  const institution = await getInstitution(ucpInstitutionId);
+
+  let aggregator: Aggregator;
+  let institutionAggregator;
+
+  if (aggregatorOverride) {
```
wesrisenmay-mx, line 53: Need a test for these changes

apps/server/src/institutions/institutionResolver.ts
```diff
@@ -35,72 +36,93 @@ const getAggregatorByVolume = (
...
 
-export async function resolveInstitutionAggregator(
-  institutionId: string,
-  jobTypes: ComboJobTypes[],
-): Promise<ResolvedInstitution> {
-  const institution = await getInstitution(institutionId);
-  const preferences = await getPreferences();
-  const aggregatorsWithAtLeastPartialSupport: Aggregator[] =
-    getAvailableAggregators({
+export async function resolveInstitutionAggregator({
```
Craiting, line 39: This function is really big and hard to follow. Could it be broken down to make it easier to read?

apps/server/src/institutions/institutionResolver.ts
```diff
@@ -35,72 +36,93 @@ const getAggregatorByVolume = (
...
+  aggregatorOverride,
+  jobTypes,
+  ucpInstitutionId,
+}: {
+  aggregatorOverride?: Aggregator;
+  jobTypes: ComboJobTypes[];
```
Craiting, line 45: jobTypes doesn't seem to be used if aggregatorOverride is passed in, should this param be optional too?

apps/server/src/institutions/getInstitutionHandler.ts
```diff
@@ -1,71 +1,37 @@
...
+  res.send({
+    guid: inst.id,
+    code: inst.id,
+    name,
     logo_url: resolvedInstitution.logo_url,
+    instructional_data: {},
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    credentials: [] as any[],
+    // This name includes thing doesn't seem like it should be here
+    supports_oauth: inst.oauth ?? name?.includes("Oauth"),
```
Craiting, line 32: I agree. I think you can rely on resolvedInstitution.oauth, especially if you update the institution resolver to return oauth based on `supports_oauth`

apps/server/src/institutions/institutionResolver.ts
```diff
@@ -35,72 +36,93 @@ const getAggregatorByVolume = (
...
+      `Resolving institution: ${ucpInstitutionId} to aggregator: ${aggregator} available aggregators: ${JSON.stringify(aggregators)}`,
+    );
   }
 
-  debug(
-    `Resolving institution: ${institutionId} to aggregator: ${aggregator} available aggregators: ${JSON.stringify(aggregators)}`,
-  );
-
   return {
     id: institutionAggregator?.id,
```
Craiting, line 127: If you add `oauth: institutionAggregator?.supports_oauth` to this object then I think you can use it in the getInstitutionsHandler

apps/server/src/institutions/getInstitutionHandler.ts
wesrisenmay-mx: Yep. Forgot about this. Good find. Fixed.

apps/server/src/institutions/institutionResolver.ts
wesrisenmay-mx: I think I'd prefer to keep it required, because it will fail if it's not provided for the non aggregator override case. The function runs fine regardless of whether aggregatorOverride is passed

apps/server/src/institutions/institutionResolver.ts
wesrisenmay-mx: Better?

apps/ui/src/postMessageEventOverrides.test.ts
```diff
@@ -0,0 +1,80 @@
...
+      it("returns true if the object has changed", () => {
+        expect(
+          postMessageEventOverrides.memberStatusUpdate.getHasStatusChanged({
+            currentMember: member1,
+            previousMember: member2,
+          }),
+        ).toBe(true);
+      });
+
+      it("returns false if the object has changed", () => {
```
Craiting, line 70: typo
```suggestion
      it("returns false if the object has not changed", () => {
```

Craiting (approved):

PR #215, opened 2025-04-10
put health check in the url

wesrisenmay-mx:


wesrisenmay-mx:
This functionality was taken care of in another PR

PR #216, opened 2025-04-10
add raw_status that's used by current clients

sophtron-jishun:
added raw_status in connection object which will later made available through client side events, the raw_status field contains the status returned from aggregator before mapped to universal widget contract, this richer information enables client to track the connection status at a more detailed and accurate level which in turns provide the opportunity to trigger more process at a better timing, as well as make diagnosing easier

this is a feature currently used by our some clients to better track the connection progress and to better diagnose.

### DIFF
--- a/apps/server/src/connect/connectApi.ts
+++ b/apps/server/src/connect/connectApi.ts
@@ -10,6 +10,7 @@ function mapConnection(connection: Connection): Member {
     institution_guid: connection.institution_code,
     guid: connection.id,
     connection_status: connection.status ?? ConnectionStatus.CREATED, // ?
+    raw_status: connection.raw_status,
     most_recent_job_guid:
       connection.status === ConnectionStatus.CONNECTED
         ? connection.cur_job_id
@@ -17,6 +18,7 @@ function mapConnection(connection: Connection): Member {
     is_oauth: connection.is_oauth,
     oauth_window_uri: connection.oauth_window_uri,
     aggregator: connection.aggregator,
+    selected_account_id: connection.selected_account_id,
     is_being_aggregated: connection.is_being_aggregated,
     user_guid: connection.userId,
     mfa: {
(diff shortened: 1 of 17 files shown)


packages/sophtron-adapter/cypress/e2e/sophtron.cy.ts
```diff
@@ -43,6 +45,40 @@ describe("Sophtron aggregator", () => {
...
+
+          [
+            "rawStatus",
+            "selectedAccountId",
+            "aggregator",
+            "member_guid",
+            "user_guid",
+            "connection_status",
+          ].forEach((prop) => {
+            expect(!!metadata[prop]).to.be.true;
```
Craiting, line 76: You can use `haveOwnProperty` here

```suggestion
            expect(metadata).to.haveOwnProperty(prop);
```

packages/sophtron-adapter/src/adapter.test.ts
```diff
@@ -488,6 +514,7 @@ describe("sophtron adapter", () => {
             JobType: "verification",
             LastStatus: accountsReadyStatus,
             UserInstitutionID: testUserInstitutionId,
+            AccounntId: "SAS_ACCOUNT_ID",
```
Craiting, line 517: Typo
```suggestion
            AccountId: "SAS_ACCOUNT_ID",
```

apps/server/src/connect/connectApi.ts
```diff
@@ -6,19 +6,40 @@ import { AggregatorAdapterBase } from "../adapters";
...
+      memberConnected: {
+        ...(connection.postMessageEventData?.memberConnected || {}),
+        ...sharedEventData,
+      },
+      memberStatusUpdate: {
+        ...(connection.postMessageEventData?.memberStatusUpdate || {}),
+        ...sharedEventData,
+        connection_status: connectionStatus,
+      },
+    },
```
Craiting, line 40: Do you want to update the internal openApi spec for this new postMessage stuff in the response?

wesrisenmay-mx: Good call

packages/sophtron-adapter/cypress/e2e/sophtron.cy.ts
Craiting: Ok makes sense.

Craiting (approved):

PR #221, opened 2025-04-29
Ct/finicity

Craiting:


packages/utils/errors.ts
```diff
@@ -1,2 +1,29 @@
 export const USER_NOT_RESOLVED_ERROR_TEXT = "Can't resolve userId";
 export const SOMETHING_WENT_WRONG_ERROR_TEXT = "Something went wrong";
+
+interface CustomError extends Error {
+  cause: {
+    statusCode?: number;
+  };
+}
+
+export const parseError = (
```
wesrisenmay-mx, line 10: As far as I can tell this is only used in apps/server, so it should live in the server's shared/utils folder.

wesrisenmay-mx, line 10: Needs unit tests

apps/server/src/connect/userEndpoints.ts
```diff
@@ -27,8 +28,11 @@ export const userDeleteHandler = withValidateAggregatorInPath(
       res.status(ret.status);
       res.send(ret.data);
     } catch (error) {
-      res.status(400);
-      res.send("User delete failed");
+      const { statusCode, message } = parseError(error);
```
wesrisenmay-mx, line 31: This function seems to be used the same way every time.

Maybe it's signature should be more like: handleError({ error, res }), and it can handle calling res.status and res.json?

packages/utils/errors.ts
```diff
@@ -1,2 +1,29 @@
...
+  error: CustomError,
+): { statusCode: number; message: string } => {
+  if (error.message === USER_NOT_RESOLVED_ERROR_TEXT) {
+    return {
+      statusCode: 404,
+      message: USER_NOT_RESOLVED_ERROR_TEXT,
+    };
+  } else if (error.message) {
+    return {
+      statusCode: error.cause?.statusCode || 400,
```
wesrisenmay-mx, line 20: Is this used anywhere as of right now?

packages/finicity-adapter/src/adapter.ts
```diff
@@ -0,0 +1,224 @@
...
+      credentials: [] as any[],
+      institution_code: request.institutionId,
+      oauth_window_uri: await this.apiClient.generateConnectLiteUrl(
+        request.institutionId,
+        userId,
+        request_id,
+      ),
+      aggregator: this.aggregator,
+      status: ConnectionStatus.CREATED,
+      jobTypes: request.jobTypes,
```
wesrisenmay-mx, line 88: As far as I can tell jobTypes isn't used anywhere except when it's retrieved from the cache to determine if it should refresh the accounts to get transactions. I don't love having it returned in all the network requests and being added to the Connection type if that's it's only use.

We could store it more like this and remove the bloat to the Connection type:

this.cacheClient.set(request_id, {
   connection: obj,
   jobTypes: request.jobTypes
})

apps/server/src/institutions/getInstitutionHandler.ts
```diff
@@ -28,7 +37,7 @@ export const getInstitutionHandler = async (req: Request, res: Response) => {
     instructional_data: {},
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
     credentials: [] as any[],
-    supports_oauth: resolvedInstitution.supportsOauth,
+    supports_oauth: oauthOverride || resolvedInstitution.supportsOauth,
```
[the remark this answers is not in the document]
wesrisenmay-mx: Yeah I think we should remove this functionality.
I recently refactored this function to solely rely on what's in the ucp institution list and ignore the previous supports oauth property that was returned from the widget adapter's getInstitutionById.

As long as the aggregator sets up their institution properly it will work, and it will be truthful about what the institution actually supports.

packages/finicity-adapter/.eslintrc.js
```diff
@@ -0,0 +1,32 @@
...
+  extends: [
+    "eslint:recommended",
+    "plugin:@typescript-eslint/recommended",
+    "prettier",
+  ],
+  ignorePatterns: [
+    ".eslintrc.js",
+    "jest.config.cjs",
+    "jestSetup.ts",
+    "cypress.config.*.js",
```
wesrisenmay-mx, line 16: This cypress stuff isn't used right?

packages/finicity-adapter/.gitignore
```diff
@@ -0,0 +1,34 @@
...
+# Test pack files
+ucp-npm-*.tgz
+
+# Logs
+logs/
+
+# IDE files
+.idea/
+
+# Environment files
```
wesrisenmay-mx, line 18: We don't need any of these env files right?

packages/finicity-adapter/.gitignore
```diff
@@ -0,0 +1,34 @@
+playwright-report
+
+# Node modules/build
+node_modules/
+dist/
+build/
+packed/
+
+# Test pack files
+ucp-npm-*.tgz
```
wesrisenmay-mx, line 10: Do we need this?

packages/finicity-adapter/jest.config.cjs
```diff
@@ -0,0 +1,15 @@
+module.exports = {
+  clearMocks: true,
+  coverageProvider: "babel",
+  collectCoverage: true,
+  collectCoverageFrom: ["src/**/*.{js,ts}"],
+  coverageDirectory: "coverage",
+  coverageReporters: ["json", "lcov", "text", "json-summary"],
+  setupFiles: ["dotenv/config"],
```
wesrisenmay-mx, line 8: Do we need this? The finicity adapter doesn't access stuff directly from process right? It all gets passed in?

packages/utils/contract.ts
```diff
@@ -157,6 +157,7 @@ export interface Connection {
   oauth_window_uri?: string | null;
   error_message?: string | null;
   postMessageEventData?: PostMessageEventData;
+  jobTypes?: ComboJobTypes[];
```
wesrisenmay-mx, line 160: Don't need this if we make the change I suggested

packages/finicity-adapter/package.json
```diff
@@ -0,0 +1,38 @@
...
+  },
+  "devDependencies": {
+    "@playwright/test": "^1.49.1",
+    "@typescript-eslint/eslint-plugin": "^6.20.0",
+    "dotenv": "^16.4.5",
+    "eslint": "^8.57.1",
+    "eslint-config-standard-with-typescript": "^43.0.1",
+    "eslint-plugin-import": "^2.29.1",
+    "eslint-plugin-jest": "^27.6.3",
+    "jest": "^30.0.0-alpha.6",
```
wesrisenmay-mx, line 32: Probably shouldn't be using an alpha version

packages/finicity-adapter/playwright/finicity.playwright.ts
```diff
@@ -0,0 +1,195 @@
...
+    await authorizeTab.getByLabel("Banking Password").fill("profile_700");
+    await authorizeTab.getByLabel("Submit").click();
+
+    const connectedPromise = new Promise<void>((resolve, reject) => {
+      const timer = setTimeout(() => reject("timed out"), 120000);
+
+      page.on("console", async (msg) => {
+        const obj = (await msg.args()[0].jsonValue())?.message;
+        if (obj?.type === "connect/memberConnected") {
+          console.log("member connected", obj);
```
wesrisenmay-mx, line 43: Do we need the log?

packages/finicity-adapter/playwright/finicity.playwright.ts
```diff
@@ -0,0 +1,195 @@
...
+  });
+
+  async function testDataEndpoints(request, userId, connectionId, aggregator) {
+    let url = `http://localhost:8080/api/data/aggregator/${aggregator}/user/${userId}/connection/${connectionId}/accounts`;
+
+    const accountsResponse = await request.get(url);
+    const accounts = await accountsResponse.json();
+    expect(accounts?.accounts?.length).toBeGreaterThanOrEqual(1);
+
+    const account = accounts.accounts[6]; // This is a deposit account
```
wesrisenmay-mx, line 112: Is there a way we could do a .find on something more resilient than an index?

packages/finicity-adapter/playwright/finicity.playwright.ts
```diff
@@ -0,0 +1,195 @@
...
+              category: expect.any(String),
+              status: "active",
+              payee: expect.any(String),
+            }),
+          }),
+        ]),
+      }),
+    );
+  }
+});
```
wesrisenmay-mx, line 195: Nice work.

packages/finicity-adapter/src/test/handlers.ts
```diff
@@ -0,0 +1,114 @@
...
+        message: "Customer not found.",
+      },
+      { status: 404 },
+    ),
+  ),
+  http.get(`${FINICITY_BASE_PATH}/aggregation/v1/customers`, ({ request }) => {
+    const url = new URL(request.url);
+    const username = url.searchParams.get("username");
+    switch (username) {
+      case "test-user-name":
```
wesrisenmay-mx, line 52: Are these cases used in several different tests? I try to just put the generic happy path in the handlers and put anything specific in the associated test file

packages/finicity-adapter/src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+      createWidgetAdapter: () =>
+        new FinicityAdapter({
+          sandbox: true,
+          dependencies,
+        }),
+    },
+  };
+};
+
+export * from "./models";
```
wesrisenmay-mx, line 32: Do these need to be exported?

packages/finicity-adapter/src/dataAdapter.ts
```diff
@@ -0,0 +1,116 @@
...
+
+        const accountTypesWithAchSupport = [
+          "checking",
+          "savings",
+          "moneyMarket",
+          "certificateOfDeposit",
+        ];
+
+        for (const account of accounts) {
+          if (accountTypesWithAchSupport.includes(account.type)) {
```
wesrisenmay-mx, line 78: Do we have a test for both these branches?

packages/finicity-adapter/src/dataAdapter.ts
```diff
@@ -0,0 +1,116 @@
...
+            accountsWithAchDetails.push({ ...account, achDetails });
+          } else {
+            accountsWithAchDetails.push(account);
+          }
+        }
+
+        return { accounts: accountsWithAchDetails.map(mapAccount) };
+      }
+      case VCDataTypes.TRANSACTIONS: {
+        const startDate = new Date(
```
wesrisenmay-mx, line 92: If we waited a month to call this endpoint, then it would no longer return data right?

packages/finicity-adapter/src/apiClient.ts
```diff
@@ -0,0 +1,237 @@
...
+      partnerId: this.apiConfig.partnerId,
+      customerId: customerId,
+      institutionId,
+      redirectUri: redir,
+      webhook: `${this.getWebhookHostUrl()}/webhook/${this.aggregator}/?connection_id=${request_id}`,
+      webhookContentType: "application/json",
+    }).then((ret: { link: string }) => ret.link);
+  }
+
+  createCustomer(unique_name: string) {
```
wesrisenmay-mx, line 166: camelCase

packages/finicity-adapter/src/apiClient.ts
```diff
@@ -0,0 +1,237 @@
...
+  }
+
+  async getAxiosInstance() {
+    if (this.axios) {
+      return this.axios;
+    }
+
+    const token = await this.getAuthToken();
+    const headers = makeFinicityAuthHeaders(this.apiConfig, token);
+    return (this.axios = axios.create({
```
wesrisenmay-mx, line 204: I'd prefer that we return it separately than shove it all in one line

packages/utils/errors.ts
Craiting: It's used on the Finicity apiClient. It allows the specific apis to send a more detailed error message rather than "Something went wrong".

Here is one example of it in action when I use the correct customer but wrong accountId: (Response status is 404)
```json
{
  "message": "customer does not have given account (customerId = 7044558470, accountId = [70793696251])"
}
```

packages/finicity-adapter/src/adapter.ts
```diff
@@ -0,0 +1,224 @@
...
+
+  async ListConnectionCredentials(
+    _connectionId: string,
+    _userId: string,
+  ): Promise<Credential[]> {
+    return [];
+  }
+
+  ListConnections(_userId: string): Promise<Connection[]> {
+    return Promise.resolve([]);
```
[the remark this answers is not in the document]
wesrisenmay-mx: I think maybe if you make it an async function it will let you do this

packages/finicity-adapter/src/adapter.test.ts
```diff
@@ -0,0 +1,381 @@
...
+        query: { connection_id: connectionId },
+      };
+
+      const updatedConnection =
+        await finicityAdapter.HandleOauthResponse(request);
+
+      expect(updatedConnection).toBeNull();
+    });
+  });
+});
```
wesrisenmay-mx, line 381: Nice work.

apps/server/src/institutions/getInstitutionHandler.ts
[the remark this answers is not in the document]
Craiting: This is the comment I meant to have here before it was reviewed but it was stuck in pending, I forgot to finish the review.

I added this override because these aggregators require the widget to use the OAuth flow. I think it's still valuable to allow individual institutions to keep a separate supports_oauth flag because Finicity does have OAuth connections with some institutions and I think it'd be good to track that in our institution service.

packages/finicity-adapter/.gitignore
Craiting: No

packages/finicity-adapter/.gitignore
Craiting: I don't think so

packages/finicity-adapter/package.json
Craiting: It turns out we don't need jest at all because it's using `ts-jest`

packages/finicity-adapter/src/apiClient.ts
```diff
@@ -0,0 +1,237 @@
...
+  constructor(
+    sandbox: boolean,
+    aggregatorCredentials: AggregatorCredentials,
+    logger: LogClient,
+    envConfig: Record<string, string>,
+    getWebhookHostUrl: () => string,
+    cacheClient: CacheClient,
+  ) {
+    this.aggregator = sandbox ? "finicity_sandbox" : "finicity";
+    (this.apiConfig = sandbox
```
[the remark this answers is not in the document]
Craiting: Oops I had a comma where a semi-colon should have been and my autoformatter did something weird

apps/server/src/institutions/getInstitutionHandler.ts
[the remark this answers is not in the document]
Craiting: I moved the oauth support check to be handled specifically by the aggregator

packages/utils/contract.ts
```diff
@@ -69,6 +69,7 @@ export interface Institution {
 export interface AggregatorInstitution {
   aggregator: string;
   id: string;
+  supports_oauth: boolean;
```
wesrisenmay-mx, line 72: camelCase

wesrisenmay-mx (approved):

PR #224, opened 2025-05-09
fix refresh connection

Craiting:


packages/finicity-adapter/src/adapter.test.ts
```diff
@@ -323,7 +409,7 @@ describe("finicity aggregator", () => {
       );
     });
 
-    it("handles 'error' reason with code 201 and updates connection status to FAILED", async () => {
+    it("handles 'error' reason with code 201 and updates connection status to Connected", async () => {
```
Craiting, line 412: It turns out code 201 means the widget exited only because there was nothing to do because the connection is already good. This is the response that comes back when you try to refresh a finicity connection which doesn't need refreshing.

packages/mx-adapter/src/adapter.ts
```diff
@@ -124,28 +125,10 @@ export class MxAdapter implements WidgetAdapter {
     return mapCredentials(res.data);
   }
 
-  async CreateConnection(
-    request: CreateConnectionRequest,
-    userId?: string,
-  ): Promise<Connection> {
+  #GetNewMemberPropertiesFromRequest(request: CreateConnectionRequest) {
```
Craiting, line 128: I learned that javascript allows private methods. The `#` makes it private. I did this because these methods still use the class properties but they are only used by other class functions. And so I don't have to write individual tests for each function, but they are still tested through the public functions which use them.

Craiting: The `CreateConnection` function was getting really big and complex so I made these other private functions in an attempt to abstract logic into more readable blocks of code.

packages/mx-adapter/tsconfig.json
```diff
@@ -3,7 +3,8 @@
     "baseUrl": "./src",
     "lib": ["ES5", "ESNext"],
 
-    "module": "ES2015",
+    "target": "ES2022",
```
Craiting, line 6: This update allowed me to use private functions

packages/mx-adapter/src/adapter.ts
```diff
@@ -190,7 +277,7 @@ export class MxAdapter implements WidgetAdapter {
         credentials: request.credentials?.map(
           (credential) =>
             ({
-              guid: credential.id,
+              guid: credential.guid || credential.id,
```
Craiting, line 280: This update allows me to call the `UpdateConnection` from `CreateConnection` without having to revert the credentials on `newMemberProperties` back to `id` and `value` after it's already been changed to guid in the `GetNewMemberPropertiesFromRequest` function.

packages/finicity-adapter/src/apiClient.ts
```diff
@@ -149,16 +149,33 @@ export default class FinicityClient {
...
       customerId: customerId,
       institutionId,
       redirectUri: redir,
-      webhook: `${this.getWebhookHostUrl()}/webhook/${this.aggregator}/?connection_id=${request_id}`,
+      webhook: `${this.getWebhookHostUrl()}/webhook/${this.aggregator}/?connection_id=${connectSessionId}`,
+      webhookContentType: "application/json",
+    }).then((ret: { link: string }) => ret.link);
+  }
+
+  generateConnectFixUrl(
```
wesrisenmay-mx, line 166: Needs a test

wesrisenmay-mx (approved):

PR #228, opened 2025-05-27
fix oauth redirect and member_guid returned

Craiting:


packages/finicity-adapter/src/adapter.ts
```diff
@@ -225,8 +226,13 @@ export class FinicityAdapter implements WidgetAdapter {
       case "credentialsUpdated":
         connection.status = ConnectionStatus.CONNECTED;
         break;
+      case "done":
+        return connection;
```
Craiting, line 230: This change fixed a problem I noticed where sometimes it seemed like there was a race between events and if this event happened first then the member_guid in the postmessage was incorrect.

packages/finicity-adapter/src/adapter.ts
```diff
@@ -225,8 +226,13 @@ export class FinicityAdapter implements WidgetAdapter {
...
       default:
         switch (reason) {
+          case "complete":
+            connection.status = ConnectionStatus.CONNECTED;
+            break;
```
Craiting, line 235: This change fixed a problem where the redirect landing page or oauth complete page was saying "Something went wrong" even though it was successful because we need to pass a successful connection status back in the connection response when the complete event type occurs.

wesrisenmay-mx (approved):
Looks good if the tests pass

PR #232, opened 2025-06-02
Ct/record performance

Craiting:


apps/server/src/institutions/getInstitutionHandler.ts
```diff
@@ -4,6 +4,7 @@ import { resolveInstitutionAggregator } from "./institutionResolver";
 
 export const getInstitutionHandler = async (req: Request, res: Response) => {
   const ucpInstitutionId = req.params.institution_guid;
+  req.context.latestResolvedInstitutionId = ucpInstitutionId;
```
Craiting, line 7: What do you think about this strategy to get the ucpInstitutionId without needing to make updates to the connect widget? In my testing it seems to be getting the institution right every time.

apps/server/src/connect/connectApi.ts
```diff
@@ -117,12 +139,29 @@ export class ConnectApi extends AggregatorAdapterBase {
...
+
+    const mainAggregatorId = getMainAggregator(this.context.aggregator);
+    const startEvent = recordStartEvent({
+      aggregatorId: mainAggregatorId,
+      connectionId: performanceSessionId,
+      institutionId: this.context.latestResolvedInstitutionId,
+      jobTypes: this.context.jobTypes,
+    });
+    if (memberData.is_oauth) {
+      // This is if we decide to include user interaction time with akoya
```
wesrisenmay-mx, line 153: If we confirm that there's no time we can measure with Akoya then we need to talk with Product/Design and figure out what makes sense for them. My guess is that their time is going to show up blank on the performance graphs and maybe an N/A in the tables

apps/server/src/connect/connectApi.ts
```diff
@@ -183,12 +222,27 @@ export class ConnectApi extends AggregatorAdapterBase {
   }
 
   async loadMemberByGuid(memberGuid: string): Promise<Member> {
-    const mfa = await this.getConnectionStatus(memberGuid);
-    if (mfa?.institution_code == null) {
+    const member = await this.getConnectionStatus(memberGuid);
+    if (member?.challenges?.length > 0) {
+      await recordConnectionPauseEvent(this.context.performanceSessionId);
```
wesrisenmay-mx, line 227: We don't want to await this right?

apps/server/src/services/performanceTracking.ts
```diff
@@ -0,0 +1,85 @@
...
+    const accessToken = await getAccessToken();
+    const url = `${config.PERFORMANCE_SERVICE_URL}/events/${connectionId}/${eventType}`;
+
+    const response = await fetch(url, {
+      method,
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+        "Content-Type": "application/json",
+      },
+      ...(body ? { body } : {}),
```
wesrisenmay-mx, line 31: You can probably just pass the body either way. My guess is passing undefined won't harm anything.

apps/server/src/connect/connectApi.ts
```diff
@@ -183,12 +222,27 @@ export class ConnectApi extends AggregatorAdapterBase {
...
+    }
+    if (member?.institution_code == null) {
       const connection = await this.getConnection(memberGuid);
-      return mapConnection({ ...mfa, ...connection });
+
+      if (member.status === ConnectionStatus.CONNECTED) {
+        if (connection.is_being_aggregated) {
+          if (connection.is_oauth) {
```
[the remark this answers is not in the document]
Craiting: The resume event for the credential flow happens in the updateMember function. It probably wouldn't make any difference to call resume here for both but it does slightly reduce load on the performance server if we limit it to only oauth.

.github/workflows/unit-tests.yml
```diff
@@ -40,6 +40,8 @@ jobs:
           echo -e "${{ env.env_vars }}" >> ${ENV_FILE_PATH}
 
           # Secrets (can't load these from another job, due to GH security features)
+          echo UCP_CLIENT_SECRET=fakeSecret >> ${ENV_FILE_PATH}
```
wesrisenmay-mx, line 43: Why do we need this?

apps/server/src/connect/connectApi.test.ts
```diff
@@ -12,22 +19,69 @@ const {
 
 const resolvedUserId = "resolvedUserId";
 
-const testContext = {
-  aggregator: MX_AGGREGATOR_STRING,
-  institutionId: "xxx",
-  resolvedUserId,
-  jobTypes: [ComboJobTypes.TRANSACTIONS],
-} as Context;
+async function waitForRequestLogLength(
```
wesrisenmay-mx, line 22: We don't have access to this function because it's node, but it would be nice to build a waitFor function that we can reuse that follows a similar API.

https://testing-library.com/docs/dom-testing-library/api-async/#waitfor

apps/server/src/connect/connectApi.test.ts
```diff
@@ -12,22 +19,69 @@ const {
...
+          },
+        ),
+        http.put(
+          `${config.PERFORMANCE_SERVICE_URL}/events/:connectionId/connectionPause`,
+          ({ params }) => {
+            requestLog.push({
+              eventType: "connectionPause",
+              connectionId: String(params.connectionId),
+            });
+            return HttpResponse.json({ ok: true });
```
wesrisenmay-mx, line 80: Wouldn't an empty object get you what you're looking for?

apps/server/src/connect/connectApi.test.ts
```diff
@@ -12,22 +19,69 @@ const {
...
+    connectApi = new ConnectApi({
+      context: testContext,
+    });
+
+    connectApi.init();
+  });
 
-describe("connectApi", () => {
   describe("addMember", () => {
     it("returns a member", async () => {
```
wesrisenmay-mx, line 54: and sends a connection start event?

apps/server/src/connect/connectApi.test.ts
```diff
@@ -12,22 +19,69 @@ const {
...
+      const requestLog: {
+        connectionId: string;
+        eventType: string;
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        body?: any;
+      }[] = [];
+      server.use(
+        http.post(
+          `${config.PERFORMANCE_SERVICE_URL}/events/:connectionId/connectionStart`,
```
wesrisenmay-mx, line 63: There's enough reuse here that it feels worth creating an abstraction for these. The function could setup the handlers and return the requestLog

packages/akoya-adapter/src/adapter.ts
```diff
@@ -154,6 +155,8 @@ export class AkoyaAdapter implements WidgetAdapter {
       return connection;
     }
 
+    this.performanceClient.recordSuccessEvent(requestId);
```
wesrisenmay-mx, line 158: How are we dealing with not being able to measure Akoya's speed?

apps/server/src/services/performanceTracking.test.ts
```diff
@@ -0,0 +1,211 @@
...
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  headers: Record<string, any>;
+}
+
+describe("performanceTracking", () => {
+  let requestLog: RequestLogEntry[] = [];
+
+  beforeEach(() => {
+    requestLog = [];
+    server.use(
```
wesrisenmay-mx, line 29: It feels a bit weird to group so many server handlers together when they're not used together in the individual tests.

.github/workflows/unit-tests.yml
Craiting: We don't really

wesrisenmay-mx (approved):

PR #241, opened 2025-06-24
Add methods in prep for performance resilience

Craiting:


apps/server/src/services/performanceResilience.ts
```diff
@@ -0,0 +1,131 @@
...
+    aggregatorId,
+    lastUiUpdateTimestamp: Date.now(),
+    pausedByMfa: false,
+  });
+};
+
+const cleanupPerformanceObject = async (
+  connectionId: string,
+): Promise<void> => {
+  await del(`${PERFORMANCE_REDIS_SUBDIRECTORY}:${connectionId}`);
```
wesrisenmay-mx, line 77: Feels like it's worth making a function that returns this string based on a provided connectionId

apps/server/src/services/performanceResilience.test.ts
```diff
@@ -0,0 +1,242 @@
...
+            },
+          }),
+        ),
+      );
+      await createPerformanceObject(basePerformanceObjectParams);
+
+      jest.useFakeTimers();
+      jest.setSystemTime(Date.now() + 8000);
+
+      const recordSuccessSpy = jest.spyOn(
```
wesrisenmay-mx, line 101: We can use MSW instead to check if the success endpoint was called

apps/server/src/services/performanceResilience.test.ts
```diff
@@ -0,0 +1,242 @@
...
+      expect(cleanedUpObject).toEqual({});
+    });
+
+    it("should do nothing if the status is PENDING", async () => {
+      server.use(
+        http.get(READ_MEMBER_STATUS_PATH, () =>
+          HttpResponse.json({
+            member: {
+              guid: "testGuid",
+              connection_status: "PENDING",
```
wesrisenmay-mx, line 125: We have an enum we can use here right?

apps/server/src/services/performanceResilience.test.ts
```diff
@@ -0,0 +1,242 @@
...
+
+      const performanceObject = await getPerformanceObject(
+        basePerformanceObjectParams.connectionId,
+      );
+      expect(performanceObject).toBeDefined();
+      expect(recordSuccessSpy).not.toHaveBeenCalled();
+    });
+
+    [
+      "IMPEDED",
```
wesrisenmay-mx, line 200: enums here as well

PR #247, opened 2025-06-30
aw/UCP-651 - add new query param

ash-wright123:
Added an new query param for an aggregator override


apps/server/src/widgetEndpoint.ts
```diff
@@ -27,6 +27,7 @@ export const widgetHandler = (req: Request, res: Response) => {
     singleAccountSelect: Joi.bool(),
     userId: Joi.string().required(),
     token: Joi.string(),
+    aggregatorOverride: Joi.string().valid(...aggregators),
```
Craiting, line 30: I think in this case it'd be better to only allow the non-sandbox aggregators names like "mx", "sophtron", "akoya". Instead of allowing "mx_int", "akoya_sand" etc. This is because the institution list which elasticsearch uses only uses the non-test agg names. I think if you had a test where aggregatorOverride is "mx_int" it wouldn't return any institutions.

apps/server/src/services/ElasticSearchClient.ts
```diff
@@ -316,13 +320,17 @@ export async function getInstitution(id: string): Promise<CachedInstitution> {
...
 }: {
   jobTypes: ComboJobTypes[];
+  aggregatorOverride?: string;
 }): Promise<CachedInstitution[]> {
   const config = getConfig();
   const preferences = await getPreferences();
 
-  const supportedAggregators = preferences.supportedAggregators;
+  const supportedAggregators = aggregatorOverride
+    ? [aggregatorOverride]
```
Craiting, line 332: If you change the joi validation to only allow prod aggregator names then this should be fine but if you want to keep the sandbox names: `mx_int`, `akoya_sandbox` then you'll probably need to resolve the official aggregator name with this function: https://github.com/Universal-Connect-Project/ucw-app/blob/main/apps/server/src/adapterIndex.ts#L6

apps/server/cypress/e2e/suite1/search.cy.ts
```diff
@@ -110,4 +110,44 @@ describe("search", () => {
       cy.findByText(MX_BANK_NAME).should("exist");
     });
   });
+
+  describe("Aggregator override", () => {
+    it("filters recommended institutions by aggregatorOverride", () => {
+      visitAgg();
+
+      const institutionThatIsInFavoriteAndSupportsAll = MX_BANK_NAME;
```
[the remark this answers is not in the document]
ash-wright123: Right and I am testing to make sure only an mx institution is shown. I made the test more clear.

apps/server/cypress/e2e/suite1/search.cy.ts
```diff
@@ -110,4 +110,47 @@ describe("search", () => {
...
   });
+
+  describe("Aggregator override", () => {
+    it("includes institutions that are only supported by MX in the recommended institution list if the aggregatorOverride is MX", () => {
+      visitAgg();
+
+      const institutionThatIsInFavoriteAndSupportsAll = MX_BANK_NAME;
+
+      cy.visit(
+        `/widget?jobTypes=${ComboJobTypes.TRANSACTIONS},${ComboJobTypes.ACCOUNT_NUMBER},${ComboJobTypes.ACCOUNT_OWNER}&userId=${crypto.randomUUID()}&aggregatorOverride=mx`,
```
wesrisenmay-mx, line 121: You could probably just pass one job type and make this test less complex

apps/server/cypress/e2e/suite1/search.cy.ts
```diff
@@ -110,4 +110,47 @@ describe("search", () => {
...
+      );
+
+      cy.findByText(institutionThatIsInFavoriteAndSupportsAll).should("exist");
+      cy.findByText(SOPHTRON_BANK_NAME, {
+        timeout: 45000,
+      }).should("not.exist");
+    });
+
+    it("does not find institutions that are only supported by other aggregators (sophtron) when aggregatorOverride is set to MX", () => {
```
wesrisenmay-mx, line 130: I'd probably combine these last 2 tests

apps/server/src/adapterSetup.ts
```diff
@@ -118,5 +118,10 @@ export const adapterMap: Record<string, AdapterMap> = {
   ...sophtronAdapterMapObject,
 };
 
+export const nonTestAggregators = Object.keys(adapterMap).filter(
+  (key) =>
```
wesrisenmay-mx, line 122: This feels fragile.

We could check all the properties of the adapter map and return a list of everything that doesn't have a testAdapterId that matches its key.

apps/server/src/widgetEndpoint.ts
```diff
@@ -48,6 +49,6 @@ export const widgetHandler = (req: Request, res: Response) => {
     path.join(__dirname, "../../ui/dist/index.html"),
     "utf8",
   );
-
+  console.log(nonTestAggregators);
```
wesrisenmay-mx, line 52: 💣

apps/server/src/institutions/getInstitutionHandler.ts
```diff
@@ -6,7 +6,8 @@ export const getInstitutionHandler = async (req: Request, res: Response) => {
   const ucpInstitutionId = req.params.institution_guid;
 
   const resolvedInstitution = await resolveInstitutionAggregator({
-    aggregatorOverride: req.context.aggregator,
+    aggregatorOverride:
+      req.context.aggregatorOverride || req.context.aggregator,
```
Craiting, line 10: Could you add a test to getInstitutionHandler.test.ts for this change. It will probably be a duplicate of another test, just with a different variable in the context

PR #249, opened 2025-07-02
issue#248 post message to proper target

sophtron-jishun:


apps/ui/src/Widget.tsx
```diff
@@ -50,10 +53,10 @@ const Widget = ({
           };
 
           if (window.parent) {
-            window.parent.postMessage(payload);
+            window.parent.postMessage(payload, refererUrl || '*');
```
emilysallstrom, line 56: This is a security vulnerability, can we not default to `'*'`? If we're opening a security window like that, I'd like it to be explicitly set by the client.

emilysallstrom (changes requested):

PR #251, opened 2025-07-11
Ct/local performance polling

Craiting:


apps/server/src/connect/connectApi.ts
```diff
@@ -198,15 +198,17 @@ export class ConnectApi extends AggregatorAdapterBase {
...
 
     if (!isRefreshConnection && memberData.is_oauth && startEvent) {
-      startEvent.then(() => recordConnectionPauseEvent(performanceSessionId));
+      startEvent.then(() =>
+        recordConnectionPauseEvent(performanceSessionId, false),
+      );
     }
 
-    if (!isRefreshConnection) {
+    if (!isRefreshConnection && this.getNeedsLocalPerformanceResilience()) {
```
wesrisenmay-mx, line 206: Why do we only need to create the performance object if it doesn't need polling?

apps/server/src/connect/connectApi.ts
```diff
@@ -198,15 +198,17 @@ export class ConnectApi extends AggregatorAdapterBase {
...
       createPerformanceObject({
         userId: this.getUserId(),
         connectionId: connection.id,
         performanceSessionId,
-        aggregatorId,
+        aggregatorId: this.context.aggregator,
```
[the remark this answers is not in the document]
Craiting: The original `aggregatorId` which was used here is the non-test aggregator id. So if the true aggregator was `mx_int` as the aggregator then the aggregatorId would be `mx`. But with the performance object it's using the aggregatorId to query the member status so it needs to separate `mx` and `mx_int` so it knows which api adapter to use. 

The original `aggregatorId` is sent to the performance service which doesn't care to separate the test environments from the production environments because it's just storing the data grouped by aggregator.

packages/finicity-adapter/src/adapter.test.ts
```diff
@@ -87,6 +87,13 @@ describe("finicity aggregator", () => {
     });
   });
 
+  describe("GetRequiresPollingForPerformance", () => {
```
wesrisenmay-mx, line 90: describe statement is no longer accurate

apps/server/src/services/performanceTracking.ts
```diff
@@ -89,19 +90,25 @@ export const recordSuccessEvent = async (connectionId: string) => {
   cleanupPerformanceObject(connectionId);
 };
 
-export const recordConnectionPauseEvent = async (connectionId: string) => {
+export const recordConnectionPauseEvent = async (
+  connectionId: string,
+  callPausePolling = true,
```
wesrisenmay-mx, line 95: I'd suggest "shouldPausePolling", but it's fine as is

wesrisenmay-mx (approved):

PR #254, opened 2025-07-21
Ct/plaid

Craiting:


packages/akoya-adapter/src/testInstitutions.ts
```diff
@@ -5,7 +5,7 @@ export const testInstitutions: CachedInstitution[] = [
     name: "Mikomo Bank",
     id: "mikomoBank",
     keywords: ["akoya", "mikomo"],
-    logo: "https://mikomo.ddp.akoya.com/img/default.png",
+    logo: "https://cdn.brandfetch.io/idwKHUTdZK/w/400/h/400/theme/dark/icon.jpeg?c=1bxid64Mup7aczewSAYMX&t=1668071436139",
```
Craiting, line 8: I went on a slight tangent and updated some test institution logos...
image

.github/workflows/e2e-playwright-plaid.yml
```diff
@@ -0,0 +1,75 @@
...
+        run: |
+          ENV_FILE_PATH=./apps/server/.env
+          touch ${ENV_FILE_PATH}
+
+          # Vars
+          echo -e "${{ needs.setup-env.outputs.env_vars }}" >> ${ENV_FILE_PATH}
+          echo PLAID_CLIENT_NAME=${{ secrets.PLAID_CLIENT_NAME }} >> ${ENV_FILE_PATH}
+          echo PLAID_CLIENT_ID=${{ secrets.PLAID_CLIENT_ID }} >> ${ENV_FILE_PATH}
+          echo PLAID_SECRET=${{ secrets.PLAID_SECRET }} >> ${ENV_FILE_PATH}
+          echo NGROK_AUTHTOKEN=${{ secrets.NGROK_AUTHTOKEN }} >> ${ENV_FILE_PATH}
```
wesrisenmay-mx, line 60: Do we need NGROK for Plaid?

.github/workflows/e2e-playwright-plaid.yml
```diff
@@ -0,0 +1,75 @@
...
+          # This fails if using https, so we have to override it
+          echo HOST_URL=http://localhost:8080 >> ${ENV_FILE_PATH}
```
wesrisenmay-mx, line 62: Do we need this for Plaid?

packages/plaid-adapter/.gitignore
```diff
@@ -0,0 +1,34 @@
...
+# Test pack files
+ucp-npm-*.tgz
+
+# Logs
+logs/
+
+# IDE files
+.idea/
+
+# Environment files
```
wesrisenmay-mx, line 18: We don't need these right?

packages/plaid-adapter/playwright.config.ts
```diff
@@ -0,0 +1,47 @@
+import { defineConfig, devices } from "@playwright/test";
+
+/**
```
wesrisenmay-mx, line 3: I don't think we need this comment

packages/plaid-adapter/src/apiClient.test.ts
```diff
@@ -0,0 +1,222 @@
+import { ComboJobTypes } from "@repo/utils";
```
wesrisenmay-mx, line 1: Nice work

packages/plaid-adapter/src/adapter.ts
```diff
@@ -0,0 +1,204 @@
...
+
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  async DeleteConnection(id: string, userId: string): Promise<void> {
+    await this.cacheClient.set(id, null);
+
+    return;
+  }
+
+  async DeleteUser(_aggregatorUserId: string): Promise<unknown> {
+    return;
```
wesrisenmay-mx, line 119: Is there no way to delete a user?

Craiting: Plaid doesn't have external users, only the access_token is required to get the data. We pass a client_user_id into the Link request but according to Plaid docs, that id is for Plaid debugging and analytics. We could invalidate the access token but we'd need to modify some documentation a little.

apps/server/src/connect/connectApi.test.ts
```diff
@@ -237,6 +237,28 @@ describe("connectApi", () => {
       );
     });
 
+    it("does not create a performance object or send a performance start event when getPerformanceEnabled returns false", async () => {
+      const requestLog = setupPerformanceHandlers([
+        "connectionStart",
+        "connectionPause",
+      ]);
+      jest.spyOn(connectApi, "getPerformanceEnabled").mockReturnValue(false);
```
wesrisenmay-mx, line 245: A more realistic test would be to get it to resolve to Plaid so that it returns false on its own.

ENVIRONMENT.md
```diff
@@ -12,9 +12,9 @@ These are the environment variables that can be set in apps/server/env/(staging|
...
-| UCP_CLIENT_ID      | Client ID, available in the [UCP dashboard](https://app.universalconnectproject.org/widget-management). Providing these api keys will sync the latest data(institutions, performance, etc.) to the widget and provide performance data to UCP. |
-| UCP_CLIENT_SECRET  | Client Secret, available in the [UCP dashboard](https://app.universalconnectproject.org/widget-management)                                                                                                                                     |
+| WEBHOOK_HOST_URL   | The webhook base url for adapters with webhooks (should be the same as HOST_URL)                                                                                                                                                               |                                    |
+| UCP_CLIENT_ID      | Client ID, available in the (UCP dashboard)[https://app.universalconnectproject.org/widget-management]. Providing these api keys will sync the latest data(institutions, performance, etc.) to the widget and provide performance data to UCP. |
```
wesrisenmay-mx, line 16: This change breaks the UCP dashboard link

packages/plaid-adapter/src/adapter.ts
```diff
@@ -0,0 +1,214 @@
...
+    await this.cacheClient.set(connectionId, null);
+    return await removeItem({
+      accessToken: connectionId,
+      clientId: this.credentials.clientId,
+      secret: this.credentials.secret,
+      sandbox: this.sandbox,
+    });
+  }
+
+  async DeleteUser(_userId: string): Promise<void> {
```
wesrisenmay-mx, line 127: Wouldn't we want to the delete user endpoint to fail, so they know to delete connection instead?

Craiting: good idea

PR #258, opened 2025-08-05
Ct/connection cleanup methods

Craiting:


apps/server/src/aggregatorPerformanceMeasuring/utils.ts
```diff
@@ -158,7 +158,6 @@ export const pollConnectionStatusIfNeeded = async (
     !connectionStatus.is_being_aggregated
   ) {
     await recordSuccessEvent(performanceSessionId);
-    cleanupPerformanceObject(performanceSessionId);
```
Craiting, line 161: this method is called in recordSuccessEvent so it's an unnecessary redundancy.

apps/server/src/connectionCleanup/utils.test.ts
```diff
@@ -0,0 +1,296 @@
...
+  });
+
+  describe("initCleanUpConnections", () => {
+    it("should throw error when cleanup is disabled", async () => {
+      jest.spyOn(config, "getConfig").mockReturnValue({
+        CONNECTION_CLEANUP_INTERVAL_MINUTES: undefined,
+        CONNECTION_CLEANUP_POLLING_INTERVAL_MINUTES: 1,
+      });
+
+      const { initCleanUpConnections } = await import("./utils");
```
wesrisenmay-mx, line 66: Are you sure you need to import it like this?

Craiting: Good catch, this is leftover from a non ideal implementation which I fixed later on. It's not necessary.

PR #261, opened 2025-08-21
Optimize performance resilience to reduce redis load

Craiting:
Pub/Sub didn't seem necessary for this after all.

### DIFF
--- a/apps/server/src/aggregatorPerformanceMeasuring/performanceResilienceManager.test.ts
+++ b/apps/server/src/aggregatorPerformanceMeasuring/performanceResilienceManager.test.ts
@@ -1,0 +1,161 @@
+import {
+  performanceResilienceManager,
+  startPerformanceResilience,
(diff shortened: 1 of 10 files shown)


apps/server/src/aggregatorPerformanceMeasuring/performanceResilienceManager.test.ts
```diff
@@ -0,0 +1,161 @@
+import {
```
wesrisenmay-mx, line 1: I feel like we're missing the most important test for this file: that this process results in performance events being recorded.

1. mock msw to return connection success
2. mock msw to watch for performance success
3. startPerformanceResilience
4. createPerformancePollingObject twice
5. wait for performance success to have been called with the correct params twice
6. expect the process to be stopped

wesrisenmay-mx (approved):
Nice work.

PR #263, opened 2025-09-05
Ct/plaid success rate

Craiting:


apps/server/src/adapters/index.ts
```diff
@@ -53,9 +53,13 @@ export class AggregatorAdapterBase {
     );
   }
 
-  async createConnection(
-    connection: CreateConnectionRequest,
-  ): Promise<Connection> {
+  async createConnection({
+    connection,
```
wesrisenmay-mx, line 57: Why not just have ucpInstitutionId be part of connection?

apps/server/src/adapters/index.test.ts
```diff
@@ -100,4 +102,127 @@ describe("AggregatorAdapterBase", () => {
       });
     });
   });
+
+  describe("createConnection", () => {
+    const mockSet = jest.spyOn(redisClient, "set");
```
wesrisenmay-mx, line 107: It would be more realistic to check if something is there than to check if a function was called

PR #271, opened 2025-10-01
Add tests around 404 response and performance resilience

Craiting:


.tool-versions
```diff
@@ -0,0 +1 @@
+nodejs 20.19.2
```
Craiting, line 1: I thought about adding this to the gitignore but it could actually be helpful to people using asdf for their version manager and I don't think it hurts anyone not using asdf.

wesrisenmay-mx (approved):

PR #272, opened 2025-10-06
Add Accounts Plaid data adapter

Craiting:


packages/plaid-adapter/playwright/utils.ts
```diff
@@ -0,0 +1,84 @@
...
+            name: "plaidInstitutionName",
+            value: expect.any(String),
+          }),
+        ]),
+      }),
+    }),
+  );
+
+  expect(accountId).not.toBeNull();
+  return accountId;
```
wesrisenmay-mx, line 83: I think you're not actually using what's returned here

Craiting: That's right but it will be used in the future when testing transactions. I can remove it until then.

wesrisenmay-mx (approved):

